Issue #6, opened 2011-12-01
ITP parser drops [dihedral] sections


orbeckst:
`[dihedral]` sections in an ITP are not handled well at the moment:
1. Multiple _subsequent_ `[dihedral]` sections are merged into a single dihedral section (e.g. proper and improper dihedrals that have been separated in the ITP file for clarity). This is only an inconvenience as it still produces the correct topology.
2. If multiple `[dihedral]` sections are _separated by another section_ then **only the last dihedral section is retained**. This is a bug as it means we can (silently!) loose data.

The second issue needs to be fixed.

Issue #23, opened 2015-10-04
unit tests framework


orbeckst:
Unit tests (with CI) would be great. Given that there are deb packages of Gromacs, one could test multiple versions of Gromacs (especially 4.6.x vs 5.0.x vs 5.1.x) but this is a lot of work at the moment.

jandom:
Hi Oli, 

I'm working on this. The best set of tests+data files I found actually belongs to pandas - I'm inspired by their structure. It's much simpler than anything I've seen (especially the datafile linking...)

My main aim will be to:
- [ ] setup the testing boilerplate
- [ ] add tests for the TOP reader I contributed

orbeckst:
When #34 is merged we will close this issue and then raise individual issues for the necessary tests.

orbeckst:
#34 was merged... so we officially have a nucleus of tests. Thanks @jandom ... onwards and upwards.

Issue #9, opened 2012-05-11
ITP #define expansion not supported 


orbeckst:
The preprocessor does not recognize #define directives that define text such as 

```
 #define dih_chi1N   -0.76567   2.70705   4.02501  -5.96639   0.00000   0.00000 ; C*     CT     CT_2   N:  Chi-1 peptides AA
```

which defines the token _dih_chi1N_ that should be expanded in lines such as 

```
   14   12   15   17     3  dih_chi1H  ; chi1 dihedral, from PHE
```

It fails with

```
SyntaxError: File: "/..../test.itp", line 187
SyntaxError: Invalid #define directive
```

The preprocessor is very simplistic and only handles simple #ifdef/#ifndef decisions and not textual replacement.

pslacerda:
Here we can use a real C preprocessor (cpp):

* clang or gcc preprocessors: http://stackoverflow.com/a/3917365/199332
* Python implementation: https://github.com/dabeaz/ply/blob/master/ply/cpp.py

I see these options:

* use current preprocessing only if gcc or clang isn't found
* use only cpp.py
* requires gcc or clang and use it

orbeckst:
I am not sure if it's worth the effort to do the parsing. The fine folks at [ParmEd](https://github.com/ParmEd/ParmEd) have figured out most of it already. 

Some rudimentary ITP handling is useful but I think GW should concentrate more on the other aspects and perhaps file formats that are not well supported elsewhere.

I am just leaving this issue open so that if anyone ever wants to use the ITP parser then at least they see what the problems are.

Issue #8, opened 2012-03-23
ITP parser does not parse .top files


orbeckst:
In principle the itp parser should also be able to read Gromacs topology files, which really are following the same syntax as itp files and are just name .top by convention (to show that they are the master file). 

At the moment, a top file is read but only the header section is available.

orbeckst:
There is now (since around 0.5.0)  a TOP parser that reads complete preprocessed top files but the current tests fail #61.

PR #25, opened 2015-12-14
Itp reader

jandom:
hi Oli, 

I've had this feature in gromacs wrapper for a long time now... it's time to push out and give back to the community. it comes with a number of tests (bash scripts and python scripts) that check both the gromacs energies (mdrun-rerun) and consistency of the itp/top files produced

Cheers, 

- Jan


gromacs/fileformats/__init__.py
```diff
@@ -4,13 +4,14 @@
 # See the file COPYING for details.
 
 # file formats
-from __future__ import absolute_import
+
```
orbeckst: please put absolute imports back in (and see below)

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
+
+
+#
+# JD: adopted from 'pytopol' from https://github.com/resal81/PyTopol
```
orbeckst: Licence compatible with GPL?

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+                return self._anumb_to_atom[anumb]
+            else:
+                print("no atoms in the molecule")
+                return False
+
+        else:
+            if anumb in self._anumb_to_atom:
+                return self._anumb_to_atom[anumb]
+            else:
+                print("no such atom number (%d) in the molecule" % (anumb))
```
orbeckst: use logger instead of print

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+        if len(self.atoms) != 0:
+
+            # reset the mapping
+            self._anumb_to_atom = {}
+
+            for i,atom in enumerate(self.atoms):
+                atom.number = i+1   # starting from 1
+
+        else:
+            print("the number of atoms is zero - no renumbering")
```
orbeckst: logger ....

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
+
+
+#
+# JD: adopted from 'pytopol' from https://github.com/resal81/PyTopol
+#
+
+class System(object):
```
orbeckst: Need some doc strings.

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+        self.interactiontypes = []
+        self.pairtypes        = []
+        self.constrainttypes  = []
+        self.forcefield=  None
+
+        self.information = {} # like 'atomtypes': self.atomtypes
+
+
+
+class Molecule(object):
```
orbeckst: doc

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
+
+#
+# JD: adopted from 'pytopol' from https://github.com/resal81/PyTopol
```
orbeckst: Need module doc string and acknowledgement of https://github.com/resal81/PyTopol

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
...
+#
+
+import logging
+from . import blocks
+from collections import OrderedDict as odict
+
+module_logger = logging.getLogger('gromacs.fileformats')
+
+
+class TOP(blocks.System):
```
orbeckst: docs

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
...
+                elif curr_sec == 'molecules':
+                    assert len(fields) == 2
+                    mname, nmol = fields[0], int(fields[1])
+
+                    # if the number of a molecule is more than 1, add copies to system.molecules
+                    for i in range(nmol):
+                        self.molecules.append(self.dict_molname_mol[mname])
+
+                else:
+                    print('Uknown section in topology: %s' % curr_sec)
```
orbeckst: typo... and use logger (and shouldn't you raise an exception??)

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
...
+        'dihedraltypes'  : '{:6s} {:6s} {:6s} {:6s}   {:1d}    {:6.2f}    {:f}    {:1d}\n',
+        'dihedrals'      : '{:3d} {:3d} {:3d} {:3d}   {:1d}\n',
+        'dihedrals_ext'  : '{:3d} {:3d} {:3d} {:3d}   {:1d}    {:6.2f}    {:f}    {:1d}\n',
+        'impropertypes'  : '{:6s} {:6s} {:6s} {:6s}   {:1d} {:6.2f} {:8.4f} \n',
+        'impropers'      : '{:3d} {:3d} {:3d} {:3d}   {:1d}\n',
+        'impropers_ext'  : '{:3d} {:3d} {:3d} {:3d}   {:1d} {:6.2f} {:8.4f} \n',
+    }
+
+
+    toptemplate = ""
```
orbeckst: that's ugly, why not use a multiline string (and possibly run through  dedent?)

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
...
+        top += '\n'
+
+        with open(self.outfile, 'w') as f:
+            f.writelines([top])
+
+        return
+        for i, m in enumerate(self.system.molecules):
+            molname = 'mol_%02d' % (i+1)
+            top += '#include "itp_%s.itp" \n' % molname
+
```
orbeckst: why all the white space?? (Generally, I'd cut down on empty lines and only do 1 line between functions/methods, 2 lines between classes)

orbeckst:
Thanks @jandom. Main issues to address
* How much of the code was taken from https://github.com/resal81/PyTopol ; I just checked and they have a  GPL v3 [LICENSE](https://github.com/resal81/PyTopol/blob/master/LICENSE) just as GW so that's fine. But you probably need to explicitly acknowledge them.
* We need **docs**, at module, class and method level. Right now it wouldn't show up in the docs and it is not clear how to use it.
* Use logger instead of `print`.
* Use absolute/relative imports under `from __future__ import absolute_import`.
* Tests are great but add a README in tests directory to say how these non-standard tests are to be run and which version of Gromacs you tested with. Having unit tests for GW would be good... see #23.

orbeckst:
Btw, does this mean we can retire the current attempt at an ITP reader and close #6, #5, #8, #9 (?) ?

jandom:
Great feedback Oli, many thanks - will get to it in the next 1-2 weeks!

jandom:
Just as a way of motivation here, for completeness. I've been playing a lot with REST2 (hamiltonian replica exchange - hrex - plumed/gromacs implementation). 

Running REST in this implementation starts with a processed.top file, which then is modified via an awk script to scale LJ, dihedrals and charges strengths. The awk script (provided by plumed manitainers) did the job but was hard to modify. Plus I knew that GW could benefit from a better itp reader. 

This may be of particular interest to people doing automated ff development with gromacs. It provides a nice python interface to handle processed.top files - doesn't work for topol.top-style files (yet).

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+
+            for i,atom in enumerate(self.atoms):
+                atom.number = i+1   # starting from 1
+
+        else:
+            print("the number of atoms is zero - no renumbering")
+
+
+class Chain(object):
+    """
```
MichalKononenko: A short description of what the class does would be useful here.

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def get_atomtype(self):
+        if hasattr(self, 'atomtype'):
+            return self.atomtype
+        else:
+            print("atom %s doesn't have atomtype" % self)
+            return False
+
+
+class Param(object):
```
MichalKononenko: doc

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+    def __init__(self):
+        self.comment = None
+        self.line  = None
+        self.disabled = False
+
+    def convert(self, reqformat):
+        assert reqformat in ('charmm', 'gromacs')
```
MichalKononenko: 1. Docstring
2. This ```if``` stack looks like it will be difficult to unit test. Could we split this into several different private methods, maybe get a Converter object going to couch these methods?

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+                    self.gromacs['param']['ljl14'] = self.charmm['param']['ljl14'] * 0.1 / (2**(1.0/6.0))
+                else:
+                    self.gromacs['param']['lje14'] = None
+                    self.gromacs['param']['ljl14'] = None
+            else:
+                raise NotImplementedError
+
+        else:
+            raise NotImplementedError
+
```
MichalKononenko: Since this is subclassed by a bunch of types, maybe defining a ``__repr__`` would help with debugging.

gromacs/fileformats/top.py
```diff
@@ -0,0 +1,1243 @@
...
+
+        return '\n'.join(data)
+
+
+
+
+    def _parse(self, fname):
+
+        _find_section = lambda line: line.strip('[').strip(']').strip()
```
MichalKononenko: This should be in its own separate ``def``.

gromacs/fileformats/blocks.py
```diff
@@ -0,0 +1,461 @@
...
+        self.impropers = []
+
+        self.cmaps     = []
+        self.pairs     = []
+        self.exclusion_numb = None  # 0, 1, 2, ..
+        self.virtual_sites3 = []
+        self.exclusions = []
+        self.settles    = []
+        self.constraints= []
+
```
MichalKononenko: Can we put this outside the constructor for ```Molecule```. Instead of

``` python

class Molecule(object)

    def __init__(self, *args, **kwargs):
        self.chains = []
```

we do

``` python
class Molecule(object)
    chains = []

    def __init__(self, *args, **kwargs):
        pass

File: gromacs/fileformats/top.py (hunk 11 above)
Comment (jandom) [inline gromacs/fileformats/top.py]: Inherited code - perfectly legible with my dyslexia "get those letters separated" :P happy to change tho

File: gromacs/fileformats/top.py (hunk 10 above)
Comment (jandom) [inline gromacs/fileformats/top.py]: that's the only comment that I don't understand, could you provide a mini example of how to replace this super-useful format dictionary lookup?

Comment (orbeckst) [inline gromacs/fileformats/top.py]: I mean the code *below* – appending to `toptemplate` just so that the indentation remains??
Why not
```python
import textwrap

   toptemplate="""
         [ defaults ]
         *DEFAULTS*
         [ atomtypes ]
         *ATOMTYPES*
         ; etc etc
         """
   toptemplate = textwrap.dedent(toptemplate)
```

jandom: Got it, thank you

gromacs/fileformats/blocks.py
jandom: I'm confused: maybe it's not clear from the code but these are instance attributes not class attributes. I don't want my Molecule instances to share `atoms` and other class members with *other* Molecule objects. 

What you're proposing will cause
```
class Molecule:
    chains = []
    def __init__(self, *args, **kwargs): 
        pass

m1 = Molecule()
m2 = Molecule()
m1.chains.append("A")
print m2.chains
Out: ['A']
```

Maybe i'm getting something wrong but this doesn't look like the desired behaviour?

jandom:
@orbeckst regarding your question: does this close issues #6, #5, #8, #9 and others. In short, no, it doesn't. The #define statement is still unsupported (neither is '#include') and this code only works for "special" top files generated by GROMACS 'grompp -pp' command. 

That said, it does provide a sane boilerplate to add support for both.

gromacs/fileformats/blocks.py
jandom: I don't understand how it would help debugging, could you explain?

jandom:
I'm inclined to ignore the rest of QuantifiedCode "checks": they're progressively anal and while true, some of them would make the code look uglier. 

As far as I'm concerned this is good to go (but not perfect of course)

orbeckst:
Quantified code is optional.

gromacs/fileformats/blocks.py
orbeckst: I agree with @jandom

orbeckst:
@jandom – also add an entry to CHANGES; just commit directly to develop

jandom:
Cheers, roger! Thank you for lightning-fast integration :) 
I'll add any new features as pull requests, rather than push directly to develop. This pull request flow works really nicely!

gromacs/fileformats/blocks.py
MichalKononenko: My apologies

orbeckst: @MichalKononenko , don't worry!!! Code review is frank and it is always very helpful to have multiple pairs of eyes. Thanks for contributing to the project!

jandom: @MichalKononenko, sorry if I came out too harsh - it's really great that
you sent your ideas

Issue #26, opened 2015-12-16
How do you make this work with gromacs 5.1.x?


dotsdl:
With gromacs 5.1 comes the disappearance of the use of gromacs tools on their own. For example,  `trjconv` now must be called with `gmx trjconv`. What changes need to happen to the library to make this work?

orbeckst:
It's a bit cumbersome but you "just" have to edit your configuration file.I just pushed out 0.4.0 and that should also update the docs.

In the tools section, describe `gmx trjconv` as
```
   gmx:trjconv
```
There are a few hard-coded hacks in there that alias new to old names (such as genbox). See 9208d52d2323b57866b2a600659cb75b400b03ff .

orbeckst:
See http://gromacswrapper.readthedocs.org/en/develop/gromacs/core/config.html#gromacs-tools-and-scripts ; took me a while to get RTD to build again. (master/latest should also be available again soon)

dotsdl:
Okay, I edited my `~/.gromacswrapper.cfg` file to include only names in the current release (5.1.1) of gromacs with the `gmx:` prefix, removing old ones. That was a bit of work...

Looks like there might be few hard-codes left in the codebase; there's at least one [here in the cookbook](https://github.com/Becksteinlab/GromacsWrapper/blob/develop/gromacs/cbook.py#L166).

Should we make this issue the focal point for making the library work from the start with gromacs 5.x only beginning with the next release? I don't think much more needs to be done (famous last words) besides finding any more hardcodes and setting the defaults in `gromacswrapper.cfg` to use the correct names.

orbeckst:
Instead of cleaning up the "hard codes" I'd actually favor a compatibility layer that allows you to either use 4.x or 5.x. That would be useful, at least for the next year or so where we are almost certainly still using 4.x from old projects and 5.x for new ones.  The [current hack in gromacs.tools](https://github.com/Becksteinlab/GromacsWrapper/blob/develop/gromacs/tools.py#L198) is not pretty at the moment but it is possible, at least for commands that only changed name. One could, of course, make a translation table and use this in a more data-driven way.

The canned command in the cook-book are a bit of a headache if the underlying G-command ceased to exist.

Overall, I agree with you that for 0.5 we should ship with 5.x defaults. A mechanism to switch between versions in the cfg file would be nice.

whitead:
[...]

PR #46, opened 2016-04-06
Added compatibility with 4.x style tool referencing

whitead:
Some scripts/canned commands rely on the 4.x program naming. `tools.G_rms` for example, which
is no longer the name of the rms gromacs command in 5.x. This creates an alias for all commands with the `G_` prefix. Like my last PR, I have no tests so am unsure if this will break any existing scripts.


orbeckst:
Good idea. Can you please 

1. add an entry to CHANGES, summarizing the PR (and reference it as PR #46)
2. add some docs to `tools.py`

orbeckst:
Come to think, one problem with this approach is that not all tools are prefixed with `g_`. For instance, `grompp`, `eneconv`, `trjconv`... should not just show up as `g_grompp`, `g_eneconv`, `g_trjconv`.

Somewhat ugly but feasible solution: create a dict with the exceptions:
```python
aliases5to4 = {
   'grompp': 'grompp',
   'eneconv': 'eneconv',
   # ...
   'sasa': 'g_sas',
   'convert_tpr': 'tpbconv',
}
try:
   alias4 = aliases4to5[name]
except KeyError:
   alias4 = "g_" + name
``` 

One could also use the reverse dict
```python
aliases4to5 = {name4: name5 for name5, name4 in aliases5to4.items()}
```

orbeckst:
Related to #26.

whitead:
I added to `CHANGES`.

I'm not sure what `tools.py` doc you would like to add. There is no discussion about a `G_` prefix in the docs. What were you looking for here?

Added code for preventing alias of a command which didn't change names and also for commands which changed names by an odd pattern (like g_sas to sasa).

gromacs/tools.py
```diff
@@ -76,14 +76,53 @@ def _generate_sphinx_class_string(clsname):
...
+}
+
 for name in sorted(config.load_tools):
     # hack for 5.x 'gmx toolname': add as gmx:toolname
     if name.startswith('gmx:'):
         name = name[4:]
+        #make alias for backwards compatability
+        #checka agianst uncommon name changes
+        #have to check each one, since it's possible there are suffixes like for double precision
+        flag = False
```
orbeckst: This is a bit ugly, can't you split off suffixes, test the stem with a dict lookup and then attach the suffix again?

And the current loop could be written
```python
# flag = False # not needed
for c5, c4 in aliases5to4.items():
    if name.startswith(c5):
        old_name = c4 + name.split(c5)[1]
        break
else:
   old_name = 'g_' + name
```

orbeckst:
For the docs I'd like to have a short description that tells the user how the names from the cfg file are translated into tool names and that we are adding compatibility names.

whitead:
I do not know how to break off the suffix. The suffix can have 0 to N
underscores. The tools also have underscores.

It sounds like you have a strong opinion about the best way to do it. I'm
just sharing my code which allows your package to be installed and
imported. You're free to rewrite this yourself to avoid ugly hacks.

The docs already say how the names are translated in the tools. I will add
a sentence about aliased names.

orbeckst:
> I do not know how to break off the suffix. The suffix can have 0 to N underscores. The tools also have underscores. 

Hm, ok, then doing `startswith()` is the way to go.

> It sounds like you have a strong opinion about the best way to do it. I'm just sharing my code which allows your package to be installed and imported. You're free to rewrite this yourself to avoid ugly hacks. 

First off, I very much appreciate you sharing your code and contributing to the project. I've been doing a fair number of code reviews and my PRs have been equally been scrutinized. My experience is that if people spend the time pointing out how code can be improved then this benefits everyone and the project itself. It is certainly not meant as a criticism of you but as a constructive criticism of a specific piece of code. And you know, I am not always right, and discussions are perfectly fine. I am sure you also have more important things to do than to fix other people's code so you can always say to me (as you did) do it yourself. But understand that I am also taking time to make sure that contributions improve the code base and sometimes I will just say I don't have the time to do this whereas the original author might be able to do this quickly and the PR hangs around.

(That was a lot of talking for a tiny thing but I wanted to make clear how I see it.)

>< The docs already say how the names are translated in the tools. I will add a sentence about aliased names.

Thanks!

You also have to rebase against develop or merge develop and address any merge conflicts before I can merge.

whitead:
Thank you for the taking the time to give feedback on my code. Your feedback has improved the PR.

This PR should not be permanent. The whole tools/config system should be rewritten to separate the python names from program names. For example, lines 179-192 in `tools.py` will not work in Gromacs 5, lines 197-198 are unnecessary, and a lack of fixed names prevents doc-strings and configuration detection.

orbeckst:
On 7 Apr, 2016, at 18:45, Andrew White wrote:

> This PR should not be permanent. The whole tools/config system should be rewritten to separate the python names from program names. For example, lines 179-192 in tools.py will not work in Gromacs 5, lines 197-198 are unnecessary, and a lack of fixed names prevents doc-strings and configuration detection.

Yes, it is in need of serious updating. Gromacs 5.x should be the focus, and support Gromacs 4.6.x when possible.

I added you as a collaborator with write access. You are now able to directly write to the repo and merge PRs.

Issue #49, opened 2016-04-08
Switch to YAML format for GromacsWrapper configs


dotsdl:
To make it possible to support multiple versions of gromacs installed on a single machine, as well as to make it easier to support custom user environments, we should switch from a INI-style config to a YAML configuration.

A portion of this config would go a long way in solving #48 and #26. For example, the schema for the tools available from various versions of gromacs could look like:

```
versions:
    4.6.5:
        serial:
            base: ""
            names:
                - grompp
                - trjconv
                - g_rms
        mpi:
            base: ""
            names:
                - mdrun_mpi

    5.1.1:
        serial:
            base: "gmx"
            names:
                - grompp
                - trjconv
                - rms
                - mdrun
        mpi:
            base: "gmx_mpi"
            names:
                - mdrun

```

We would then probably require specifying which version one plans to use on import. So, one can do:

``` python
import gromacs
gromacs.listconfig()             # could use interactively to list config keys available
gromacs.useconfig('4.6.5')  # this will build the classes from the tools found

```

This does not solve the problem where the same name is present multiple times within the same config, such as for `mdrun` in `5.1.1` above. At present the same class names are built regardless of the version used, so this would result in an exception. 

Unless we build class names with clear namespace differences (such as using the base given in the config), we can't get around this problem. Although possible, having a library that changes its API depending on the config is probably not the best of ideas.

dotsdl:
I may poke at this this weekend, since I'm gearing up to use GromacsWrapper extensively again.

orbeckst:
As we just discussed, definitely a step in the right direction, I'd very much like to see something like this. It's going to be a bit of a hassle to convert existing cfg files but a user only has to do this once (and maybe we can even cook up a script to do this when the time comes). 

For the name collisions I would just raise an error and have the user make the names unique within one version. (We can later think of other ways to make it work, e.g. allowing classes to carry "tags" around so that you can provide an additional kwarg to select the version of `mdrun` that you configured. there's still the problem that you already mentioned that the API is suddenly defined in the configuration file but on the other hand this is already true because the pre- and postfixes of Gromacs commands are fully user configurable....) 

I really like the idea of doing 
```python
import gromacs
```
and not having any gromacs tools loaded (which currently incurs a significant performance penalty because each tool is run to check it's available and to extract its doc string). When `import gromacs` is cheap it's easier to use the other functionality such as the different fileformat parsers.

[...]

Issue #39, opened 2016-02-26
set up CI


orbeckst:
Once we have a tests infrastructure in place #23, we need
- [x] Travis CI
- [x] coverage
- [x] add test and coverage badges to README

orbeckst:
For most of the tests we need to install Gromacs as well. This *should* be possible through `aptitude` on the containerized Travis infrastructure.

* Gromacs 4.6.5 is available as [trusty (14.04 LTS): gromacs (4.6.5-1build1) [universe]](http://packages.ubuntu.com/trusty/gromacs)
* Gromacs 5.0.x: [wily: gromacs (5.0.6-1) [universe] ](http://packages.ubuntu.com/source/wily/gromacs)
* Gromacs 5.1.x: [Xenial: gromacs (5.1.2-1ubuntu1) [universe]](http://packages.ubuntu.com/source/xenial/gromacs)

I don't know if it's possible to pull in newer packages into a 14.04 machine. Gromacs 4.6.5 will be easy, all the other... we will see. It would be great, though, to have a test matrix with different Gromacs installations.

dotsdl:
We can use conda. There are builds provided by acellera: https://anaconda.org/acellera/gromacs

orbeckst:
@dotsdl can you add the travis stuff from MDPOW? This seems to be *almost* working.

dotsdl:
Yes. I'll convene with @ianmkenney and we'll slay this one.

dotsdl:
We now have working CI, but some of the tests are failing. Haven't dug into fixing these yet; check the test output for starting points: https://travis-ci.org/Becksteinlab/GromacsWrapper

Coverage reads at 88%. Not bad, but I guess we cheat a bit since we generate many classes dynamically.

orbeckst:
@dotsdl : I don't see the coverage at https://codecov.io/gh/Becksteinlab/GromacsWrapper — is this because `codecov` is only ran when the tests pass?

And the [output from travis](https://travis-ci.org/Becksteinlab/GromacsWrapper#L343) seems to indicate that the 88% coverage are the coverage *in the test files*, not in the code itself.

I added the badges to the README but until coverage shows up in codecov I will leave this issue open.

dotsdl:
@orbeckst ah sorry, didn't look closely enough at what was included in coverage. I'll fix this.

dotsdl:
[...]

Issue #32, opened 2016-01-19
docs incomplete on readthedocs


orbeckst:
There are huge gaps in the autogenerated docs on http://gromacswrapper.readthedocs.org. Needs to be read, checked, and fixed.

Example: missing docs for most building block modules: http://gromacswrapper.readthedocs.org/en/latest/gromacs/building_blocks.html

dotsdl:
We can use [conda](http://docs.readthedocs.org/en/latest/conda.html) to build environment on readthedocs now. We use this for [MDSynthesis](http://mdsynthesis.readthedocs.org/) and it works quite well. Happy to set this up.

There are conda gromacs builds from acellera we can use to grab a recent gromacs to build classes against: https://anaconda.org/acellera

orbeckst:
Please try it!

dotsdl:
Okay, we have docs that build the environment, including gromacs 5.1, using conda: https://readthedocs.org/projects/mdsynthesis/builds/3881150/

But, when building the class docs for the building blocks, we get errors because the config we ship with is for gromacs 4.x, and so things like ``grompp`` can't be found. Quick and dirty solution is to replace current [gromacswrapper.cfg](https://github.com/Becksteinlab/GromacsWrapper/blob/develop/gromacs/templates/gromacswrapper.cfg) with one that works with 5.x names. Longer-term solution is #49.

We cannot install a gromacs 4.x version on readthedocs with conda because there are are no 4.x packages available on the anaconda cloud. If we want this, we'd have to build our own first.

orbeckst:
> 
> We cannot install a gromacs 4.x version on readthedocs with conda because there are are no 4.x packages available on the anaconda cloud. If we want this, we’d have to build our own first.

There are Ubuntu packages that we can use for 4.x … probably in parallel to condo packages.

--
Oliver Beckstein * orbeckst@gmx.net
skype: orbeckst  * orbeckst@gmail.com

dotsdl:
@orbeckst how would we do this under readthedocs? We can't use ``apt`` to build the environment on readthedocs.

orbeckst:
> On 13 Apr, 2016, at 16:06, David Dotson <notifications@github.com> wrote:
> 
> @orbeckst how would we do this under readthedocs? We can’t use apt to build the environment on readthedocs.

You’re right, of course. I was thinking about Travis… 

If we can have 5.x on RTD then perhaps we should just ship with 5.x cfg.

--
[...]

PR #53, opened 2016-04-20
Updated config to be compatible with Gromacs 5.1.1

dotsdl:
Since Gromacs 5.1.x removed the use of g_* methods, instead opting only
for the new ``gmx <command>`` style, this config features the necessary
bits to use the commands this way. This configuration will *not* work
for Gromacs 4.x or earlier. We will hopefully replace this config style
with one that works for multiple versions more flexibly with #49.

Removed ``extras``, as these tools are no longer maintained and it's not
clear that they will work as expected under this config.


dotsdl:
With this merge, we should got proper doc builds (#32), and it will go towards getting CI up and running under travis (#39).

orbeckst:
fastest merge ever ;-)

PR #55, opened 2016-05-14
Tries to detect and load GMXRC automatically

pslacerda:
This commit enables to load GMXRC automatically in most common cases. If GMXPRFIX variable is not found in the environmnet, GMXRC is ran in a subprocess and the variables exported by it are set in this environment. I do exactly the same in my weekend toy project [pslacerda/gmx](https://github.com/pslacerda/gmx). I hope it helps!

### DIFF
--- a/gromacs/__init__.py
+++ b/gromacs/__init__.py
@@ -242,6 +242,17 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+# Try to load environment variables setted by GMXRC
+gmxrc = config.cfg.get("Gromacs", "GMXRC")
+if gmxrc:
+    try:
+        config.set_gmxrc_environment(gmxrc)
+    except OSError:
+        logger = config.logger
+        logger.warning("Failed to automatically set the Gromacs environment"
+                       "(GMXRC)")
+
+
 # Add gromacs command **instances** to the top level.
 # These serve as the equivalence of running commands in the shell.
 # (Note that each gromacs command is actually run when the instance is
--- a/gromacs/config.py
+++ b/gromacs/config.py
@@ -166,6 +166,11 @@ command ``gmx`` is added as a prefix::
    # Release of the Gromacs package to which information in this sections applies.
    release = 5.0.5
 
+   # GMXRC contains the path for GMXRC file which will be loaded. If not
+   provided is expected that it was sourced as usual before importing this
+   library.
+   GMXRC = /usr/local/gromacs/bin/GMXRC
+
    # tools contains the command names of all Gromacs tools for which classes are generated.
    # Editing this list has only an effect when the package is reloaded.
    # (Note that this example has a much shorter list than the actual default.)
@@ -215,7 +220,7 @@ completely transparent to the user.
 """
 from __future__ import absolute_import, with_statement
 
-import os, errno
+import os, errno, subprocess
 from ConfigParser import SafeConfigParser
 
 from pkg_resources import resource_filename, resource_listdir
@@ -480,6 +485,7 @@ class GMXConfigParser(SafeConfigParser):
           self.set('DEFAULT', 'managerdir',
                   os.path.join("%(configdir)s", os.path.basename(defaults['managerdir'])))
           self.add_section('Gromacs')
+          self.set("Gromacs", "GMXRC", "")
           self.set("Gromacs", "tools", "pdb2gmx editconf grompp genbox genion mdrun trjcat trjconv")
           self.set("Gromacs", "extra", "")
           self.set("Gromacs", "groups", "tools")
@@ -637,9 +643,24 @@ def check_setup():
 check_setup()
 
 
-
 # Gromacs tools
 # -------------
+
+#: Runs GMXRC in a subprocess and put environment variables loaded by it in
+#: this environment.
+def set_gmxrc_environment(gmxrc):
+    envvars = ['GMXPREFIX', 'GMXBIN', 'GMXLDLIB', 'GMXMAN', 'GMXDATA',
+               'GROMACS_DIR', 'LD_LIBRARY_PATH', 'MANPATH', 'PKG_CONFIG_PATH',
+               'GROMACS_DIR', 'PATH']
+    cmdargs = ['bash', '-c', ". {0} && echo {1}".format(gmxrc,
+               ' '.join(['${0}'.format(v) for v in envvars]))]
+    try:
+        out = subprocess.check_output(cmdargs)
+        out = out.strip().split()
+        for key, value in zip(envvars, out):
+            os.environ[key] = value
+    except (subprocess.CalledProcessError, OSError):
+        pass
 
 #: Python list of all tool file names. Filled from values in the tool
 #: groups in the configuration file.
--- a/gromacs/cbook.py
+++ b/gromacs/cbook.py
@@ -200,11 +200,11 @@ one to fit.
 try:
     _define_canned_commands()
 except (OSError, ImportError, GromacsError):
-    warnings.warn("Failed to define a number of commands in gromacs.cbook. Most likely the "
-                  "Gromacs installation cannot be found --- source GMXRC!",
-                  category=GromacsImportWarning)
-    logger.error("Failed to define a number of commands in gromacs.cbook. Most likely the "
-                  "Gromacs installation cannot be found --- source GMXRC!")
+    msg = ("Failed to define a number of commands in gromacs.cbook. Most "
+          "likely the Gromacs installation cannot be found --- set GMXRC in "
+           "~/.gromacswrapper.cfg or source GMXRC directly")
+    warnings.warn(msg, category=GromacsImportWarning)
+    logger.error(msg)
 finally:
     del _define_canned_commands
 


pslacerda:
The exception handling failed check does not apply because is the intended behavior. The one about string interpolation does not mean too much either yet because there are other similar construct in the file.

gromacs/__init__.py
```diff
@@ -242,6 +242,28 @@ def stop_logging():
...
 
+
+# Check if GMXRC was loaded and tries to load it otherwise. Runs GMXRC in a
+# subprocess and put environment variables loaded by it in this environment.
+import os, subprocess
+if 'GMXPREFIX' not in os.environ:
+    envvars = ['GMXPREFIX', 'GMXBIN', 'GMXLDLIB', 'GMXMAN', 'GMXDATA',
+               'GROMACS_DIR', 'LD_LIBRARY_PATH', 'MANPATH', 'PKG_CONFIG_PATH',
+               'GROMACS_DIR', 'PATH']
+    cmdargs = ['bash', '-c', ". /usr/local/gromacs/bin/GMXRC && echo %s" %
```
orbeckst: Don't hard-code the path to GMXRC. In `bash` you can also source anything in your `PATH`, so try just
```python
cmdargs = ['bash', '-c', ". GMXRC && echo %s" %
```

orbeckst: Might as well start using modern string formatting...

gromacs/__init__.py
```diff
@@ -242,6 +242,28 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+
+# Check if GMXRC was loaded and tries to load it otherwise. Runs GMXRC in a
```
orbeckst: I'd like to make this a function. Then we can add some logic to only run it if the user wants it, we can test it separately, and we can wrap it in anice try/except.

Put it into `config.py` and call the function something like
```python
def set_gmxrc_environment(GMXRC="GMXRC"):

```

orbeckst:
The QC things are not a big deal but see the comments: let's make it a function and then call it from `__init__` as 
```python

import .config

try:
   config.set_gmxrc_environment()
except OSError:
   logger.warning("Failed to automatically set the Gromacs environment (GMXRC)")
```

orbeckst:
(Btw, @pslacerda I added a link to your [gmxscript](https://github.com/pslacerda/gmx) to the [Alternatives](http://gromacswrapper.readthedocs.io/en/latest/alternatives.html) section.)

pslacerda:
Do you mean something like the updated commit?
This may break testing if is to be done using other GMXRC than default that wasn't sourced before running it.

And many thanks by the mention in your page!

pslacerda:
@orbeckst, I'd prefer to simply set the environment in `__init__.py` rather than put in a separate function at `config.py` because the tools wouldn't be reloaded even importing the module again.

Maybe move the first `for` loop in `tools.py` into a function to enable reloading the tools on demand? But why? So I'd prefer to keep it directly in `__init__.py`.

gromacs/__init__.py
```diff
@@ -242,6 +242,15 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+# Try to load environment variables setted by GMXRC
+try:
+    import os
+    if 'GMXPREFIX' not in os.environ:
```
orbeckst: `GMXPREFIX` only works for Gromacs 5.x; we don't have it for Gromacs 4.6.x. Check for the presence of `GMXBIN` instead (or first `GMXPREFIX` and then `GMXBIN`).

orbeckst:
> Do you mean something like the updated commit?

That looks ok for a start. I don't think it will break anything at the moment because you sensibly check if Gromacs was loaded (however, as I said in the comments above, you need to check for `GMXBIN` and not just `GMXPREFIX`). Thus, if one loads the Gromacs environment before starting Python then everything will work as before.

What I don't like is that at the moment the "auto loading" will only work if your GMXRC is `GMXRC='/usr/local/gromacs/bin/GMXRC'`. 

But we could make this configurable in the config file:

```ini
[Gromacs]

GMXRC = /usr/local/gromacs/bin/GMXRC

tools = ...
```

and then use the config file value for `GMXRC`. Especially once we have #49 implemented and have ways to switch between different profiles, this could be really useful.

What do you think?

> Maybe move the first for loop in tools.py into a function to enable reloading the tools on demand?

That's a possibility but we can address this in another PR.

pslacerda:
Hi @orbeckst!

I put a `GMXRC` variable in the `GMXConfigParser` as you suggested. If it's empty `__init__.py` will fail as before otherwise it'll `set_gmxrc_environment()`.

Please check if the documentation is appropriate, if not how can I improve it?

pslacerda:
I guess that version 4 soon will be used only for reproduce old experiments, if ever used. Drop support for version 4 will greatly improve the code. What do you think about?

If better decide to maintain both versions, couldn't version 4 native commands be located more automatically? For example, in 5 we can query the command list with `gmx help commands`. In 4 the binaries are all in the same directory and, if I remember correctly, all starts with `g_` so a `glob.glob('/path/g_*')` will give us all available commands. Leaving manual configuration for home made commands, which I never needed but knows that some people do.

Also if `GMXRC` is already set, there is yet any need to `release` configuration option?

Great library, I'm starting to use it now and seems that it will be very helpful in my new project!

orbeckst:
> On 30 May, 2016, at 17:10, Pedro Lacerda <notifications@github.com> wrote:
> 
> I guess that version 4 soon will be used only for reproduce old experiments, if ever used. Drop support for version 4 will greatly improve the code. What do you think about?

I think you’re correct in your assessment but it is useful to be able to run 4.6.x for a while. Yes, it’s a bit cluncky but convenient and to be honest, the main reason for GW is convenience… it is certainly not elegant code, it can break, etc…. but in my experience it still beats bash scripts ;-)

The GW code is not as sleek as your gmxscript – perhaps we could simplify GW somewhat along the lines what you did.


> If better decide to maintain both versions, couldn't version 4 native commands be located more automatically? For example, in 5 we can query the command list with gmx help commands. In 4 the binaries are all in the same directory and, if I remember correctly, all starts with g_ so a glob.glob('/path/g_*') will give us all available commands. Leaving manual configuration for home made commands, which I never needed but knows that some people do.

I like that idea (I think there was already a discussion in one of the issues). I saw that you parse ‘gmx help’.

> Also if GMXRC is already set, there is yet any need to release configuration option?

Not really.

The release config option is not used at the moment. It had been added as a reminder. I like keeping meta data. 

> Great library, I'm starting to use it now and seems that it will be very helpful in my new project!

I am happy to hear this, especially because gmxscript already looks like a pretty neat little tool in itself. On that note, GW is a really open (not to say, promiscuous or wild west) open source project – if you have ideas then submit PRs! I tend to review them with an eye towards “is this going to break scripts horribly or not” (because with my PI hat on I really hate anyone spending time on rewriting existing code in a scientifically unproductive  manner) but otherwise, this is very much user driven!

gromacs/__init__.py
```diff
@@ -242,6 +242,17 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+# Try to load environment variables setted by GMXRC
+gmxrc = config.cfg.get("Gromacs", "GMXRC")
+if gmxrc != "":
```
orbeckst: Just make this
```
if not gmxrc:
```

Generally better to just test for ``False`` because then one can also set it to ``None`` or ``False``.

orbeckst:
1. Why would this not work with Gromacs 4.x (there's a comment stating it)? Are there any env vars that differ? Did you test with Gromacs 4?
2. Please fix the small stylistic test.

Otherwise looking good.

(Sorry for the delay in looking at this. I have so many other things to do and working with code is more a reward when I completed lots of other boring stuff... In the future, feel free to add a comment "ping @orbeckst, I think this is ready..." or something like that as a reminder.)

pslacerda:
I really don't know why I said that will not work in 4.x, maybe because wasn't tested.

Just fixed the if test, I think this is ready!
Sorry for the delay, I'm not receiving updates in my email!

pslacerda:
Let's continue this conversation about code simplification in #49. I'm really glad that you like gmxscript, and even more if some tricks did there could be applied here!

gromacs/__init__.py
```diff
@@ -242,6 +242,17 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+# Try to load environment variables setted by GMXRC
+gmxrc = config.cfg.get("Gromacs", "GMXRC")
+if not gmxrc:
```
orbeckst: @pslacerda , I am sorry, I am a total idiot asking for the change `if gmxrc != ""` to `if not gmxrc:`: This is now doing the *opposite*. Instead it should be
```python
if gmxrc:
```

D'oh! I will fix it.

And come to think: If the user has an old cfg file, the `cfg.get()` will through a `NoOptionError`. But I will fix this by adding Gromacs/GMXRC to the defaults.

orbeckst: ... Ignore my last comment regarding the `cfg.get()`, I just realized you had already added a default of "".

gromacs/config.py
```diff
@@ -637,10 +643,25 @@ def check_setup():
...
+               'GROMACS_DIR', 'LD_LIBRARY_PATH', 'MANPATH', 'PKG_CONFIG_PATH',
+               'GROMACS_DIR', 'PATH']
+    cmdargs = ['bash', '-c', ". {0} && echo {1}".format(gmxrc,
+               ' '.join(['${0}'.format(v) for v in envvars]))]
+    try:
+        out = subprocess.check_output(cmdargs)
+        out = out.strip().split()
+        for key, value in zip(envvars, out):
+            os.environ[key] = value
+    except subprocess.CalledProcessError, OSError:
```
orbeckst: in Py2 you need parentheses:
```python
except (subprocess.CalledProcessError, OSError):
```

otherwise the error trace will be assigned to the variable `OSError` and it will *not* trap `OSError`.

pslacerda:
And I'm another because didn't saw that!
Also fixed the parentheses!

pslacerda:
What can I say... tomorrow I'll change a little bit the code because now I guess that it will break in Python versions pre 3.5 because of `subproces.check_output()`.

Or is it ok for a new release?

orbeckst:
Python 2.7 is fine with [subprocess.check_output()](https://docs.python.org/2/library/subprocess.html#subprocess.check_output).

orbeckst:
I added to the PR but because I can't commit to yours I am closing this one in favor of #66.

I'll also give you write access so that you can work on the PR directly.

pslacerda:
Maybe can we probe the /usr/local/gromacs/bin/GMXRC just in the common case of default compilation and installation? No side effect if it fails as we handle like no GMXRC was provided.

PR #57, opened 2016-05-24
Made prefixes per-tool customizable

whitead:
Now can do things like prefix gmx_mpi:mdrun and gmx:grompp and the commands will execute using the different prefixes (called drivers in code). Addresses Issue #48

PR #58, opened 2016-05-24
Made prefixes per-tool customizable - Redo

whitead:
See discussion from PR #57. Started developing on same branch and added extra crap. Github gets confused if I try rebasing on a branch that is in pending PR, so I had to swtich


orbeckst:
I manually tested with 4.6.5 and seemed to work. Thanks!

PR #67, opened 2016-06-29
MDRunner can select appropriate mdrun for either Gromacs 4.x or 5.x

### LINKED ISSUE #64: Gromacs 5.x MDrunner class implementation incompatibilities
I have found two errors in regards to the MDrunner class and the `check_mdrun_success` function found in `run.py`.

1. Command passed to subprocess in the class method `run` does not include `gmx`, resulting in an `OSError`
2. `check_mdrun_success` searches the last couple of lines of the log file for the string `"Finished mdrun on node"` at the start of the line, while a successful run writes `"Finished mdrun on rank"`

My temporary fixes to these were as follows:
1. 
```python
cmd = self.commandline(**mpiargs) 
cmd = ['gmx'] + cmd
```
But this is clearly a sloppy solution while breaking Gromacs 4.x usage
2. Simply change `if line.startswith("Finished mdrun on node")` to `if line.startswith("Finished mdrun on rank")`

orbeckst:
- fixes #64
- added a hack whereby multiple commands can be supplied in the MDRunner.mdrun
  attribute: the new run.find_gromacs_command() function tries them in order
  and takes the first one that seems to work
- added utilities.which() + test

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ whitead, dotsdl, orbeckst
 
 * fixed: check_mdrun_success() works now for Gromacs 4 and Gromacs 5
   (issue #64)
+* fixed: MDRunner working for Gromacs 4 and Gromacs 5 (issue #64)  
 * fixed setup.energy_minimize() not falling back to single precision
   mdrun (issue #63)
 * fixed: added missing alias "gmx solvate" <--> "genbox" (issue #62)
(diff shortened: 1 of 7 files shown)


orbeckst:
This has not been tested yet and is work in progress. Feedback welcome.

orbeckst:
@ianmkenney please test if you find the time.

orbeckst:
Fixed PR and amended commit and force-pushed.

orbeckst:
This now working as far as I can tell (and I added a test). I will merge manually and then do the 0.5.1 release as discussed in #70.

orbeckst:
This was merged manually and rebased.

Issue #68, opened 2016-06-29
Automatic tool loading


pslacerda:
I suggest to load the tools automatically as show in the untested code below. It uses `suffix` and `version` options if present. Custom commands can be injected using the `extra`. We can also move the `GromacsCommandMultiIndex._fake_multi_ndx` logic to a function `merge_ndx()` so users can use it like `grompp(n=merge_ndx('site1.ndx', 'site2.ndx')` in any tool they wants.

```python
def load_v5_tools():
    gmx_cmd = append_suffix('gmx')
    try:
        out = subprocess.check_output([gmx_cmd, '-quiet', 'help', 'commands'])
    except subprocess.CalledProcessError:
        raise GromacsToolLoadingError("Failed to load v5 tools")

    registry = {}
    for line in str(out, 'ascii').splitlines()[5:-2]:
        if line[4] != ' ':
            name = line[4:line.index(' ', 4)]
            fancy = name.replace('-', '_')
            registry[fancy] = type(fancy, (GromacsCommand,),
                                   {'command_name':name})
    return registry


def load_v4_tools():
    tools = ("do_dssp", "editconf", "eneconv", "g_anadock", "g_anaeig",
             "g_analyze", "g_angle", "g_bar", "g_bond", "g_bundle", "g_chi",
             "g_cluster", "g_clustsize", "g_confrms", "g_covar", "g_current",
             "g_density", "g_densmap", "g_densorder", "g_dielectric",
             "g_dipoles", "g_disre", "g_dist", "g_dos", "g_dyecoupl",
             "g_dyndom", "genbox", "genconf", "g_enemat", "g_energy",
             "genion", "genrestr", "g_filter", "g_gyrate", "g_h2order",
             "g_hbond", "g_helix", "g_helixorient", "g_hydorder",
             "g_kinetics", "g_lie", "g_luck", "g_mdmat", "g_membed",
             "g_mindist", "g_morph", "g_msd", "gmxcheck", "gmxdump",
             "g_nmeig", "g_nmens", "g_nmtraj", "g_options",
             "g_order", "g_pme_error", "g_polystat", "g_potential",
             "g_principal", "g_protonate", "g_rama", "g_rdf", "g_rms",
             "g_rmsdist", "g_rmsf", "grompp", "g_rotacf", "g_rotmat",
             "g_saltbr", "g_sans", "g_sas", "g_select", "g_sgangle",
[...]

PR #66, opened 2016-06-09
discover and load GMXRC

orbeckst:
- can provide GMXRC in cfg file and load environment inside the Python environment
- replaces PR #55 by @pslacerda

Issue #61, opened 2016-05-25
top reader is mostly broken


orbeckst:
Not all of the current tests pass. Need to either fix bugs or adjust tests.

orbeckst:
@jandom can you have a look – you wrote most of these tests.

With Gromacs 4.6.5, I get the following
```
(develop) yngvi:temp oliver$ nosetests -v ../gromacs/
gromacs.tests.test_basic.test_empty ... ok
gromacs.tests.test_example.test_empty ... ok
test_fileformats.test_fileformats ... ok
test_amber03star.TestCharmm.test_basic ... ok
Load the same topology twice and check if __eq__ comparisions work ... ok
Check if grompp can be run successfully at all ... ok
Check if grompp can be run successfully at all ... ERROR
Called by `test_parameters()` for each molecule in the system ... ok
Test if all the parameter types are the same across two topologies ... ok
Test if per-molecule parameters and parameter type assignments are identical ... ok
Read a topology, write it out, and read in the output again. ... ERROR
test_amber03w.TestCharmm.test_basic ... ok
Load the same topology twice and check if __eq__ comparisions work ... ok
Check if grompp can be run successfully at all ... ERROR
Check if grompp can be run successfully at all ... ERROR
Called by `test_parameters()` for each molecule in the system ... ok
Test if all the parameter types are the same across two topologies ... ok
Test if per-molecule parameters and parameter type assignments are identical ... ok
Read a topology, write it out, and read in the output again. ... ERROR
test_charmm22.TestCharmm.test_basic ... ok
Load the same topology twice and check if __eq__ comparisions work ... ok
Check if grompp can be run successfully at all ... ok
Check if grompp can be run successfully at all ... /Volumes/Data/oliver/Biop/Projects/Methods/GromacsWrapper/gromacs/tests/test_fileformats/test_top/top.py:153: FutureWarning: sort(....) is deprecated, use sort_index(.....)
  assert_frame_equal(df1.sort(axis=1), df2.sort(axis=1), check_names=True)
/Volumes/Data/oliver/Biop/Projects/Methods/GromacsWrapper/gromacs/tests/test_fileformats/test_top/top.py:165: FutureWarning: sort(....) is deprecated, use sort_index(.....)
  assert_frame_equal(df1[unscaled_terms].sort(axis=1), df3[unscaled_terms].sort(axis=1), check_names=True)
[...]

Issue #56, opened 2016-05-23
release 0.5.0


orbeckst:
Get out a 0.5.0 release so that we can use it from PyPi.

Will push the CI and tests to a 0.5.1.

orbeckst:
I'll push out a 0.5.0. We'll do a 0.5.1 for fixes, docs, and tests.

orbeckst:
GromacsWrapper 0.5.0 released with packages on pypi.

Thanks everyone.

PR #69, opened 2016-06-29
Load tools automatically

pslacerda:
Proposal for tool loading as in issue #68. Awaiting for opinions or changes proposals.

Even if almost complete it remains largely untested (eg. `extra` and `groups`).


pslacerda:
How can the following code of `GromacsCommandMultiIndex` works if `ndx`is a list?

```python
# concatenate all index files
make_ndx = Make_ndx(f=kwargs.get('s'), n=ndx)
rc,out,err = make_ndx(o=self.multi_ndx, input=['q'], stdout=False, stderr=False)
```

Probably worked in some previous version. So can I assume that is safe to remove `GromacsCommandMultiIndex`?

pslacerda:
Sorry, I was wrong. Of course it works!
Seems that now this PR is fully backward compatible. Please also take a look at `merge_ndx()`.

pslacerda:
Please look this PR! Of course it will not be accepted as is, it just has some ideas that may be useful.

For example `load_v5_tools()`  ignoring the `tools` option because they can be fully automatically detected. Also I think that `set_gmx_environment()` must raise an error when the `gmxrc` argument value is invalid.

I apologize if my issues and commits seems disorganized because git just don't fit in my head! My local repo is messy and I do a lot of funny stuff to push a commit.

gromacs/tools.py
```diff
@@ -94,175 +71,202 @@ def _generate_sphinx_class_string(clsname):
     'mdrun': 'mdrun',
     'make_ndx': 'make_ndx',
     'make_edi': 'make_edi',
-    'gmxdump': 'gmxdump',
-    'gmxcheck': 'gmxcheck',
+    'dump': 'gmxdump',
```
orbeckst: This looks like a separate fix. Could you please make it a separate PR?

gromacs/__init__.py
```diff
@@ -242,6 +242,12 @@ def stop_logging():
     log.clear_handlers(logger)  # this _should_ do the job...
 
 
+# Try to load environment variables set by GMXRC
```
orbeckst: Please rebase against develop – I had to rewrite history for the 0.5.1 release #56  in develop and this probably messed up your PR. Sorry.

It looks to me that you're trying to merge again the GMXRC loading PR #66.

gromacs/config.py
```diff
@@ -125,13 +125,46 @@
 Gromacs tools and scripts
 -------------------------
 
-``load_*`` variables are lists that contain instructions to other
-parts of the code which packages and scripts should be wrapped.
+Fundamentally, GromacsWrapper makes existing Gromacs tools
```
orbeckst: All of this was in PR #66 – rebase to develop.

orbeckst:
HI @pslacerda , can you please rebase to develop? During the 0.5.1 release #70 I had to rewrite the history of develop.

I haven't had time to do a full review of your PR (and I also want to see what is really new compared to develop). I like that you created more functions. Ping me when you're done, sorry, a bit busy at the moment!

gromacs/tools.py
pslacerda: Done #71!

pslacerda:
Of all that code seems that `load_v5_tools()` and `merge_ndx()` looks almost ready. The `extra` and `tools` options seems to work as expected. The `suffix` option was introduced to apply the suffix at v5 driver or v4 commands. If the `tools` option isn't provided, v4 guesses will be made from `TOOLS_V4`. I'm not sure if all v4 installations provides `g_luck` or if it's better to test `mdrun` or `pdb2mgx`. I'm also not sure if `GromacsCommand.__doc__` is being properly set.

pslacerda:
Now it has more adherence to the project's programming style. The `driver` or `suffix` option is needed to support double precision on Gromacs 5.x. Double precision isn't supported for 5.x, neither extraction of docstring by running commands with `-h`.

gromacs/config.py
```diff
@@ -724,9 +724,18 @@ def set_gmxrc_environment(gmxrc):
...
                        "from GMXRC=%r", gmxrc)
 
-#: Python list of all tool file names. Filled from values in the tool
-#: groups in the configuration file.
-load_tools = []
-for g in cfg.getlist('Gromacs', 'groups', sort=False):
-     load_tools.extend(cfg.getlist('Gromacs', g))
-del g
+
+def get_tools():
```
orbeckst: Good idea to make it a function.

gromacs/config.py
```diff
@@ -724,9 +724,18 @@ def set_gmxrc_environment(gmxrc):
...
+
+    """
+    load_tools = []
+    for g in cfg.getlist('Gromacs', 'groups', sort=False):
+         load_tools.extend(cfg.getlist('Gromacs', g, sort=False))
+    return load_tools
+
+
+RELEASE = cfg.get('Gromacs', 'release')
+MAJOR_RELEASE = RELEASE.split('.')[0]
```
orbeckst: What do you want to use the `MAJOR_RELEASE` for?

It's not a bad idea but at the moment, `release` is not actually required in the cfg file. At a minimum we should allow something like `5.1.*` or even `5.*.*` so that users don't have to declare every patch-level release of Gromacs.

Furthermore, if Gromacs changes their release numbering scheme then we have to change this, too.

For right now, put this in a try/except and if getting a `RELEASE` fails, just set everything to `None` (or to 4 or 5, depending on the guessing elsewhere?)

gromacs/tools.py
```diff
@@ -59,27 +59,13 @@
 from __future__ import absolute_import
 __docformat__ = "restructuredtext en"
 
-import os.path
-import tempfile
+import os.path, tempfile, subprocess, atexit, warnings
```
orbeckst: PEP8 style is [one line per import](https://www.python.org/dev/peps/pep-0008/#imports), please adhere to it.

gromacs/tools.py
```diff
@@ -59,27 +59,13 @@
...
 
-from . import config
-from .core import GromacsCommand, Command
-from . import utilities
+from . import config, exceptions
```
orbeckst: one per line

gromacs/tools.py
```diff
@@ -103,168 +89,155 @@ def _generate_sphinx_class_string(clsname):
...
-                                                '__doc__': "Gromacs tool '%(prefix) %(name)r'." % vars()})
-        #add alias for old name
-        #No need to see if old_name == name since we'll just clobber the item in registry
-        registry[old_clsname] = cls
+
+def tool_factory(clsname, name):
+    """ GromacsCommand derived type factory. """
+    return type(clsname, (GromacsCommand,), {
+        'command_name': name,
+        'driver': 'gmx' if config.MAJOR_RELEASE == '5' else None
```
orbeckst: In this way users loose the freedom to mix single (`gmx:mdrun`) and double precision (`gmx_d:mdrun`) tools. 

I don't like loosing this ability, especially, because the old code below is handling this case already.

orbeckst: *If* we want to use `MAJOR_RELEASE` then better make it an argument of `tool_factory` instead of accessing `config` inside.

gromacs/tools.py
```diff
@@ -103,168 +89,155 @@ def _generate_sphinx_class_string(clsname):
...
+
+    :return: dict mapping tool names to GromacsCommand classes
+    """
+    tools = {}
+    for name in config.get_tools():
+        fancy = name.capitalize().replace('-', '_')
+        tools[fancy] = tool_factory(fancy, name)
+    try:
+        null = open(os.devnull, 'w')
+        subprocess.check_call(['g_luck'], stdout=null, stderr=null)
```
orbeckst: I'd rather use `grompp` as a proxy tool.

gromacs/tools.py
```diff
@@ -103,168 +89,155 @@ def _generate_sphinx_class_string(clsname):
...
+    """ Load Gromacs 4.x tools.
+
+    :return: dict mapping tool names to GromacsCommand classes
+    """
+    tools = {}
+    for name in config.get_tools():
+        fancy = name.capitalize().replace('-', '_')
+        tools[fancy] = tool_factory(fancy, name)
+    try:
+        null = open(os.devnull, 'w')
```
orbeckst: good idea

gromacs/tools.py
```diff
@@ -103,168 +89,155 @@ def _generate_sphinx_class_string(clsname):
...
+def tool_factory(clsname, name):
+    """ GromacsCommand derived type factory. """
+    return type(clsname, (GromacsCommand,), {
+        'command_name': name,
+        'driver': 'gmx' if config.MAJOR_RELEASE == '5' else None
+    })
+
+
+def load_v5_tools():
+    """ Load Gromacs 5.x tools.
```
orbeckst: Docs should say that it *automatically* infers all tools from running `gmx help`.

gromacs/tools.py
orbeckst: In fact, it might be useful to add some code to `Command` so that the special device name '/dev/null' "just works".

gromacs/tools.py
```diff
@@ -103,168 +89,155 @@ def _generate_sphinx_class_string(clsname):
...
+    return tools
+
+
+def load_v4_tools():
+    """ Load Gromacs 4.x tools.
+
+    :return: dict mapping tool names to GromacsCommand classes
+    """
+    tools = {}
+    for name in config.get_tools():
```
orbeckst: `config.get_tools` does not automatically find tools. See https://github.com/Becksteinlab/GromacsWrapper/issues/49#issuecomment-224136177 for one possibility for how to do autodetection,

orbeckst:
Hi @pslacerda,

sorry for the long silence.

Good ideas and I like packing code into functions.

Can you read through my Summary below and check if understand correctly what is happening?

### Summary
As I understand it at the moment:
- The cfg file *must* now specify the **release**.
- Gromacs 5 
  - tools are autodetected by parsing `gmx help` — this relies on `GMXRC` having been parsed already: is this guaranteed?
  - The driver command is always `gmx`; in particular, it is not possible to use `gmx_d` or mix in e.g. `gmx:mdrun` (which currently becomes `gromacs.tools.Mdrun`) and `gmx_d:mdrun` (which becomes `gromacs.tools.Mdrun_d`).
- There are no more `gmx:toolname` entries for Gromacs 5 in the cfg file.
- Gromacs 4 tools are read from the *Gromacs.tools* section of the cfg file and are not autodetected.

### Current concerns
1. I don't know why the doc strings do not work for Gromacs 5 tools but that should be fixed so that it works. 
2. I also don't want to give up the flexibility to use single and double precision and, say, MPI binaries all in one session.
3. Is it still possible to set the tools for Gromacs 5 in the cfg file?
4. `extras` is not used often but it is convenient in some cases, so I wouldn't want to give it up unless it becomes a major headache.

### Discussion
I think we need a good idea what the config file should look like (discussion in #49). This PR suggests that we should have
* required: Gromacs release
* optional: path to `GMXRC` (for autoloading without manual sourcing of `GMXRC`)
* ... what else?

Would it be possible to break this PR in two pieces:
1. simple refactoring (functions etc) and *optional* autoloading of Gromacs 5 and 4 tools (e.g., if the `Gromacs.tools` section of the cfg file is empty or set to some magic such as `__AUTOLOAD__`.
2. changes that mess with the cfg file format (e.g., getting rid of Gromacs 5 `gmx:toolname` entries).

pslacerda:
By default version 5 tools is used and release has no impact on it. I'll think better and comment when arrive at home.

With the rethinking in #49 may be easier to put even different kinds of binaries in the same group.

orbeckst:
Pedro, I'm traveling (to SciPy 2016, hooray) so might be a bit limited in when I can look at things so please have some patience. Thanks. 

--
Oliver Beckstein
email: orbeckst@gmail.com

pslacerda:
Very nice feedback, I'm very happy to have a chance to collaborate with this project! Have a great SciPy!

pslacerda:
Hi @orbeckst ,

I tried to put in code some of your observations. Now I'm separating the refactoring part of this PR into a separated commit.

### Current modifications

* Moved class and functions to the beginning of module.
* Removed the need of a release specified.
* Now `GMXRC` is optional as you suggested. If it isn't provided, the code will fall back graciously, but will really fail if no tools was detected.
* Gromacs 5 are loaded automatically using some heuristics and can also be configured like this:
```ini
[Gromacs]
GMXRC = /path/to/GMXRC
tools = gmx gmx_d gmx_mpi gmx_mpi_d
```
* Gromacs 4 tools remains untested but are also loaded using similar heuristic, but my guess that it is very slow to load.
* I have no idea to what to do with the `extra` options and I agree that is important. When someone make custom commands in Gromacs 5 it appears listed on `gmx help` or it is compiled into a separated binary?
* Still doesn't understand the `Convert_tpr`/`Tpbconv` 5.0.5 hack.

Cheers!

pslacerda:
I also don't understand the need to `trjconv` to `trjconv`, but is really needed!

pslacerda:
Reading a bit more I understood why there are tool name aliased to itself and added a comment.

This commit is only for fixes and documentation reorganization.

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+        names = find_executables(os.environ['GMXBIN'])
+
+    if len(names) == 0 or len(names) > len(V4TOOLS) * 4:
+        names = V4TOOLS[:]
+
+    tools = {}
+    for name in names:
+        try:
+            null = open(os.devnull, 'w')
+            subprocess.check_call([name], stdout=null, stderr=null)
```
orbeckst: Why are we trying to execute every single gromacs 4 command here?

And each command will fail because it would need at least `-v` or `-h`. — Or maybe I misunderstand the code here.

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+        except subprocess.CalledProcessError:
+            pass
+        fancy = make_valid_identifier(name)
```
orbeckst: Just create the tool and we fail lazily later once a user tries to access the tool.

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+def load_extra_tools():
+    """ Load extra tools.
+
+    :return: dict mapping tool names to GromacsCommand classes
+    """
+    names = config.cfg.get("Gromacs", "extra").split()
+    tools = {}
+    for name in names:
+        fancy = make_valid_identifier(name)
+        driver = 'gmx' if config.MAJOR_RELEASE == '5' else None
```
orbeckst: Hard-coding the driver here limits the extra tools. However, I don't really have a good idea how to make this better short of using the `gmx:toolname` notation (which also allows e.g. `gmx_d:toolname`). We could go with this solution, **document it**, and if someone needs more control, raise an issue and work on it.

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+    :param args: index files and zero or one structure file
+    :return: path for the new merged index file
+    """
+    ndxs = []
+    struct = None
+    for fname in args:
+        if fname.endswith('.ndx'):
+            ndxs.append(fname)
+        else:
+            assert struct is None, "only one structure file supported"
```
orbeckst: Don't use `assert` to check user input – one cannot easily work with it in user code and when run with `python -O`, asseryions are optimized away. Just raise an exception with `ValueError`.

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+    __doc__ += ".. class:: %s\n    :noindex:\n" % name
 
-#  and clean up the module scope
-cls = clsname = name = rec = doc = None  # make sure they exist, because the next line
-del rec, name, cls, clsname, doc  # would throw NameError if no tool was configured
 
+# finally add command classes to module's scope
+globals().update(registry)
+__all__ = ['GromacsCommandMultiIndex', 'merge_ndx']
+__all__ = __all__ + registry.keys()
```
orbeckst:
```python
__all__.extend(registry.keys())
```

gromacs/tools.py
```diff
@@ -94,177 +108,262 @@ def _generate_sphinx_class_string(clsname):
...
+            'driver': registry[name5].driver,
+            '__doc__': registry[name5].__doc__
+        })
+        registry[name4] = klass
+        if name5 in registry == '5':
+            registry[name5] = klass
 
 
 # 5.0.5 compatibility hack
 if 'Convert_tpr' in registry:
```
orbeckst: Stupid question... I can't remember why we are doing this here. Can't we just include that in the aliases?

orbeckst:
Looking pretty good but I have to try it locally.

gromacs/tools.py
pslacerda: Sure, `-h` is needed here.

gromacs/tools.py
pslacerda: This would increase the importing times in about a half. Currently loading times is to call all the commands by forking it (50%) and acquiring the help (50%). Here we aren't communicating for grab the help. If #76 is accepted, both modifications together would make the lib load in about a half of the time, but only if for v4 tools are used.

gromacs/tools.py
pslacerda: We would allow mixed version tools or this makes no sense?

gromacs/tools.py
pslacerda: You're right!

pslacerda:
Seems that all default options defined in `templates/gromacswrapper.cfg` except `release` are already filled by default in the `config` module (`configdir`, `qscriptdir` and so on).

`tools` and `groups` aren't really needed any more. Yeah! Maybe for v4 just as precaution because the tool loading is a bit more sensitive. Maybe we can remove this now almost obsolete file! Or maybe both files `templates/gromacswrapper.cfg` and `templates/gromacswrapper_465.cfg`.

gromacs/tools.py
```diff
@@ -94,177 +108,244 @@ def _generate_sphinx_class_string(clsname):
...
+    'convert_tpr': 'tpbconv',
+    'dump': 'gmxdump',
+    'check': 'gmxcheck',
     'solvate': 'genbox',
+    'distance': 'g_dist',
+    'sasa': 'g_sas',
+    'gangle': 'g_sgangle',
+
+    # 5.0.5 compatibility hack
+    'tpbconv': 'convert_tpr'
```
orbeckst: I think I mentioned previously something along the lines "oh, this line could be merged with the aliases". But when looking at it in context, this seems dumb because we already have
```python
 'convert_tpr': 'tpbconv',
```

I am not really sure what's even happening now – does `convert_tpr` disappear?

I *think* we can just completely remove the line. The original motivation was that 5.0.x still shipped with `g_*` tools but they were inconsistent and did not include `tpbconv` anymore but had rather renamed it to `convert_tpr`. The hack predated the aliasing system. 

Let's remove the line and if there are any bug reports, we will address them...

Sorry about that.

gromacs/tools.py
```diff
@@ -94,177 +108,244 @@ def _generate_sphinx_class_string(clsname):
...
-    # dynamically build the module doc string
-    __doc__ += _generate_sphinx_class_string(clsname)
-
-# modify/fix classes as necessary
-# Note:
-# - check if class was defined in first place
-# - replace class
-# - update local context AND registry as done below
+class GromacsToolLoadingError(Exception):
+    """Raised when could not find any Gromacs command."""
```
orbeckst: Raised when no Gromacs tool could be found.

gromacs/tools.py
```diff
@@ -94,177 +108,244 @@ def _generate_sphinx_class_string(clsname):
...
+                kwargs['n'] = merge_ndx(*ndx)
+        return kwargs
+
+
+def tool_factory(clsname, name, driver, base=GromacsCommand):
+    """ Factory for GromacsCommand derived types. """
+    clsdict = {
+        'command_name': name,
+        'driver': driver,
+        '__doc__': property(base._get_gmx_docs)
```
orbeckst: Still loving it :-).

gromacs/tools.py
```diff
@@ -94,177 +108,244 @@ def _generate_sphinx_class_string(clsname):
...
+                         'demux.pl', 'xplor2gmx.pl']):
+            execs.append(exe)
+    return execs
+
+
+def load_v5_tools():
+    """ Load Gromacs 5.x tools automatically using some heuristic.
+
+    Tries to load tools (1) using the driver from configured groups (2) and
+    fails back to automatic detection from ``GMXBIN`` (3) then to rough guesses.
```
orbeckst: falls back

gromacs/tools.py
```diff
@@ -94,177 +108,244 @@ def _generate_sphinx_class_string(clsname):
...
+    struct = None
+    for fname in args:
+        if fname.endswith('.ndx'):
+            ndxs.append(fname)
+        else:
+            if struct is not None:
+                raise ValueError("only one structure file supported")
+            struct = fname
+
+    fd, multi_ndx = tempfile.mkstemp(suffix='.ndx', prefix='multi_')
```
orbeckst: Need a line
```python
os.close(fd)
```
Otherwise we leak file descriptors.

orbeckst:
At this stage, documentation becomes important:

* update/add docs that describe what the cfg file should look like for
   * Gromacs 5.x
   * Gromacs 4.x
* add docs that briefly outline how GromacsWrapper is loading tools; in particular in which order do things happen, what's the role of the cfg file and GMXRC or an already loaded Gromacs environment. In principle, all of this is already somewhere in the doc strings and in the comments in this PR but it should be bundled in one place so that the users and developers can quickly get an overview. In fact, I suggest we make this a whole new page in the docs: `doc/sphinx/source/gromacs/configuration.txt` and reference it in [doc/sphinx/source/gromacs.txt](https://github.com/Becksteinlab/GromacsWrapper/blob/develop/doc/sphinx/source/gromacs.txt) in the `toctree` section:
   ```rest
   .. toctree::
       :maxdepth: 2

       gromacs/init
       gromacs/configuration
       gromacs/core
       gromacs/building_blocks
   ```

* the docs in `config.py` should be checked so that they don't contain any wrong information; however, they can be shortened because the `configuration.txt` docs should be the main source of information for users. The config docs themselves can then be reduced to stuff for developers.
* entry to CHANGES (referencing #68)

I'll do a bit more testing and let you know if I find anything else.

We can then also discuss if we can get rid of the template files.

gromacs/tools.py
pslacerda: Very nice!

gromacs/tools.py
pslacerda: Done

pslacerda:
Hi, just created `configuration.txt`, again it needs review. It's missing reduct `config.py` and edit CHANGES, I'll do it soon.

pslacerda:
@orbeckst, I appreciate the idea of having a repository of .mdp files. It may be very helpful.

Maybe not the ones we are using right now. I don't have many in hand as I'm not doing many simulations these days (mostly a programming guy). The .mdp options keep changing and some old files I have are already somewhat outdated.

There isn't so many kinds of .mdp out there. We may create a site where users can adjust the relevant options and simply copy & paste into their projects. Power users will probably not need it, except if they are lazy (which many are), but many Gromacs users may find it helpful to not keep a personal .mdp repository themselves or recreate the files every time. Not speaking about users that know what to do but have no idea on how to do.

What is your opinion?

gromacs/config.py
```diff
@@ -222,14 +201,6 @@
 
    gromacs.grompp(f="md.mdp", c="system.gro", p="topol.top", o="md.tpr")
 
-(The driver command is stripped and only the "command name" is used to
```
orbeckst: The new `configuration.txt` file should also say under which names tools appear. Eg if a driver is `gmx` then it will generate `gromacs.mdrun` from `gmx mdrun` but if the driver is `gmx_d` the this will be `gromacs.mdrun_d` (I think... *I* would like to see the docs myself).

doc/sphinx/source/configuration.txt
```diff
@@ -0,0 +1,106 @@
...
+program.
+
+As there isn't yet any way to know which Gromacs version to use,
+**GromacsWrapper** will first try to use Gromacs 5 if available, then to use
+Gromacs 4. If you have both versions and want's to use version 4 or just wants
+to document it, you may specify the which release version will be used::
+
+  [Gromacs]
+  GMXRC = /usr/local/gromacs/bin/GMXRC
+  release = 4.7
```
orbeckst: There won't be 4.7. The latest is 4.6.7

doc/sphinx/source/configuration.txt
```diff
@@ -0,0 +1,106 @@
...
+
+For now **GromacsWrapper** will guess which tools are available to put it into
+:mod:`gromacs.tools`, but you can always configure it manually. Gromacs 5 has
+up to 4 commands usually named::
+
+  [Gromacs]
+  tools = gmx gmx_d gmx_mpi gmx_mpi_d
+
```
orbeckst: Possibly say hear how the commands will appear in GW, see comment below.

gromacs/config.py
```diff
@@ -266,50 +228,36 @@
...
 from pkg_resources import resource_filename, resource_listdir
 
 from . import utilities
 
-# Defaults
-# --------
-# hard-coded package defaults
 
-#: name of the global configuration file.
+# Default name of the global configuration file.
```
orbeckst: Leave the colon: this will make sphinx generate docs for `CONFIGNAME`:
```python
#: Default name ...
CONFIGNAME = ...
```

gromacs/config.py
```diff
@@ -266,50 +228,36 @@
...
-#:    used by :mod:`gromacs.qsub`.
-#:    The default value is ``~/.gromacswrapper/qscripts``.
-#: :data:`templatesdir`
-#:    Directory to store user supplied template files such as mdp files.
-#:    The default value is ``~/.gromacswrapper/templates``.
-#: :data:`managerdir`
-#:    Directory to store configuration files for different queuing system
-#:    managers as used in :mod:`gromacs.manager`.
-#:    The default value is ``~/.gromacswrapper/managers``.
+# More default values
```
orbeckst: Why did you remove the docs? They were in the generated docs and the information is not available elsewhere, I think.

gromacs/config.py
```diff
@@ -372,7 +317,7 @@ def _generate_template_dict(dirname):
     by external code. All template filenames are stored in
     :data:`config.templates`.
     """
-    return dict((resource_basename(fn), resource_filename(__name__, dirname+'/'+fn))
+    return dict((resource_basename(fn), resource_filename(__name__, dirname + '/' + fn))
```
orbeckst: If you are correcting spaces, just correct it to use `os.path.join()` – that's what I should have used in the first place.

gromacs/config.py
```diff
@@ -722,3 +652,21 @@ def get_tool_names():
...
+
+
+RELEASE = None
+MAJOR_RELEASE = None
+
+if cfg.get('Gromacs', 'release'):
+    RELEASE = cfg.get('Gromacs', 'release')
+    MAJOR_RELEASE = RELEASE.split('.')[0]
+
+check_setup()
```
orbeckst: add newline

gromacs/templates/gromacswrapper.cfg
```diff
@@ -21,19 +21,7 @@ release = 5.1.1
...
-	gmx:eneconv     gmx:covar       gmx:filter      gmx:morph      gmx:rdf      gmx:sigeps      gmx:solvate             gmx:pdb2gmx
-	gmx:anadock     gmx:current     gmx:gyrate      gmx:msd        gmx:sorient  gmx:genconf     gmx:insert-molecules
-	gmx:anaeig      gmx:density     gmx:h2order     gmx:nmeig      gmx:rms      gmx:spatial     gmx:genion      
-	gmx:analyze     gmx:densmap     gmx:hbond       gmx:nmens      gmx:rmsdist  gmx:spol        gmx:genrestr    
-	gmx:angle       gmx:dielectric  gmx:helix       gmx:nmtraj     gmx:rmsf     gmx:tcaf        gmx:check       
-	gmx:bar         gmx:helixorient gmx:order       gmx:rotacf     gmx:traj     gmx:dump        gmx:trjorder
-	gmx:dipoles     gmx:kinetics    gmx:pme_error   gmx:rotmat     gmx:tune_pme gmx:grompp      gmx:mk_angndx
-	gmx:bundle      gmx:disre       gmx:lie         gmx:polystat   gmx:saltbr   gmx:vanhove     gmx:make_edi    
-	gmx:chi         gmx:distance    gmx:potential   gmx:sas        gmx:velacc   gmx:make_ndx    gmx:xpm2ps
+tools = gmx gmx_d
```
orbeckst: This breaks the existing file format. Can we add a check somewhere that raises an error (`ValueError`?) if we detect entries with `gmx[^:]*:.*` and then say that they need to change their input file, see URL to new configuration docs?

orbeckst: By the way, such a check doesn't have to be in this PR – raise an issue so that we add something before the next release.

doc/sphinx/source/configuration.txt
```diff
@@ -0,0 +1,106 @@
+==============
+ Configuration
+==============
+
```
orbeckst: add
```rest
.. versionchanged:: 0.6.0
   The format of the ``tools`` variable in the ``[Gromacs]`` section of the config file was changed for Gromacs 5 commands.
```

orbeckst:
Good changes, see comments above, but I think we're almost there.

gromacs/config.py
pslacerda: Because is developer documentation quickly implied by variable names, except `qscriptdir` which again can be found just searching the project.

orbeckst: I don't like deleting documentation that is still correct. It's quite a lot of work to write it in the first place an I'm always grateful when I find variables properly documented. Unlike code comments it shows tht someone thought about how to communicate to a user or another developer and not just to oneself. 

--
Oliver Beckstein
email: orbeckst@gmail.com


> Am Jul 26, 2016 um 16:44 schrieb Pedro Lacerda <notifications@github.com>:
> 
> Because is developer documentation quickly implied by variable names, except qscriptdir which again can be found just searching the project.

pslacerda:
ping @orbeckst,
All fixes were done. A review of configuration.txt will also improve my English writing.

pslacerda:
What do you think about my previous suggestion of a mdp file dynamic repository. When the user know what he is doing is more a matter of join together mdp options and adjust parameters. Maybe we can help people at doing this. Molecular dynamics for noobs and repository for the rest, along with previously made GromacsWrapper programs we can seriously help to popularize molecular simulations.

pslacerda:
Also the commits of this PR needs to be combined into one so don't mess with the repository history. And the patch-0.5.1 branch can already be deleted as was already merged into develop.

I'm not very skilled at git and github but seems relatively important things to do.

Following this PR I have some patches for core.py like this one, would be my last major patch at hand. I also propose move the plugins into a separated repository, will be cleaner and make easier to move just the main code to Python 3, for example.

pslacerda:
Yes, seems that we're almost there with this PR!

pslacerda:
Probably also remove or update the branch gh-pages
http://becksteinlab.github.io/GromacsWrapper/

orbeckst:
@pslacerda – sorry for the delay... working on some other urgent stuff. Please ping again if nothing happened by Wednesday.

orbeckst:
@pslacerda , quick comments while H ave 5 minutes to spare:

1. For the discussion on an mdp repository, please open a new issue and flag it with the *question* tag (just copy and paste  https://github.com/Becksteinlab/GromacsWrapper/pull/69#issuecomment-234733706 and  https://github.com/Becksteinlab/GromacsWrapper/pull/69#issuecomment-235771482 ) – so that we can keep the discussion focused.

2. I also like the idea of splitting off the plugins https://github.com/Becksteinlab/GromacsWrapper/pull/69#issuecomment-235772400 and make the core library leaner. Please open a new issue.

Regarding this PR: I can do some rebasing and history massaging. I just need to test locally. I'll also look at the docs... in fact, I will probably get rid of the gh-pages branch, given that rtd http://gromacswrapper.readthedocs.io/ works for us – I opened #80.

orbeckst:
Sorry @pslacerda – still not managed to try it. Busy with urgent stuff. Next week should be better.

orbeckst:
With the new changes, one cannot install Gromacswrapper if Gromacs has not been sourced. See the output on travis job [#102.1](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/152529235) and
```
(auto) yngvi:GromacsWrapper oliver$ pip install --upgrade .
Processing /Volumes/Data/oliver/Biop/Projects/Methods/GromacsWrapper
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/var/folders/qs/k07z49fh8xl6xd008k8wxhxr0000gp/T/pip-YFccaC-build/setup.py", line 17, in <module>
        version = __import__('gromacs.version').get_version()
      File "gromacs/__init__.py", line 252, in <module>
        from . import tools
      File "gromacs/tools.py", line 296, in <module>
        registry = load_v5_tools()
      File "gromacs/tools.py", line 230, in load_v5_tools
        raise GromacsToolLoadingError("Failed to load v5 tools")
    gromacs.tools.GromacsToolLoadingError: Failed to load v5 tools

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /var/folders/qs/k07z49fh8xl6xd008k8wxhxr0000gp/T/pip-YFccaC-build/
```

If I source `GMXRC` before installing then it works without problems.

However, we need to be able to install without Gromacs present.

pslacerda:
I didn't had this problem here but now the setup.py doesn't import the whole package before install it, just import the version file.

pslacerda:
I'm not understanding the errors on [#103.1][1] as they doesn't happen here when running `py.test gromacs`. What do you think?
screen

[1]: https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/152831127

orbeckst:
Sorry @pslacerda , this took forever. I ironed out a few bugs related to Gromacs v4 loading but seems to be good now. Merging now....

orbeckst:
regarding errors on travis: not sure, @ianmkenney is working on tests for MDPOW and anything that he finds/cooks up will also be applicable to GW.

You can always raise an issue for tests that don't behave. I think there's already #61 (and #47).

pslacerda:
Very nice @orbeckst! Was very rich experience commit on this project, I learned a lot about refactoring in Python. :grinning:

For sure I completly forgot about v4, thank you for fixing it.

What [this line](https://github.com/Becksteinlab/GromacsWrapper/pull/69/commits/e10893478c7a4ed7a5f775540934cab124a1f0a4#diff-1c9d51497d63dbee10437e60a7727c1cR650) is doing? Doesn't it breaks the environment variables placing `__` around it?

gromacs/config.py
```diff
@@ -639,11 +639,15 @@ def set_gmxrc_environment(gmxrc):
...
+    # only v5: 'GMXPREFIX', 'GROMACS_DIR'
+    envvars = ['GMXBIN', 'GMXLDLIB', 'GMXMAN', 'GMXDATA',
+               'LD_LIBRARY_PATH', 'MANPATH', 'PKG_CONFIG_PATH',
+               'PATH',
+               'GMXPREFIX', 'GROMACS_DIR']
+    # in order to keep empty values, add ___ sentinels around result
+    # (will be removed later)
     cmdargs = ['bash', '-c', ". {0} && echo {1}".format(gmxrc,
-               ' '.join(['${0}'.format(v) for v in envvars]))]
+               ' '.join(['___${{{0}}}___'.format(v) for v in envvars]))]
```
orbeckst: If a env var such as `GMXPREFIX` or `GROMACS_DIR` is not defined then it produces an empty value. Empty values are not picked up when the output is parsed and lead to wrong assignments of values to env vars. That was a problem with Gromacs 4.

Moving unknown env vars to the end would have been sufficient but using "Sentinel" markers around will ensure that even an empty result is picked up because it produces `___'___` (the ' should not be there but indicates where "nothing" is). The line below then removes the sentinels. Cludgy but fairly safe.

(Of course, everything just breaks if the env var values contain spaces as in `GMXBIN=/Users/joe/Stuff for work/gromacs`...)

orbeckst: Btw, it might be better to put the values on new lines and then split on those:
```python
cmdargs = ['bash', '-c', ". {0} && echo {1}".format(gmxrc, 
           '\n'.join(['${{{0}}}'.format(v) for v in envvars]))]
...
    out = out.split('\n')
    for key, value in zip(envvars, out):
             value = value.strip()
             os.environ[key] = value
```

This will work even for subsequent empty lines:
```python
>>> out = "\n".join(["line 1", "line 2", "", "line 4", "", "", "line_5", ""])
>>> out.split("\n")
['line 1', 'line 2', '', 'line 4', '', '', 'line_5', '']
```

pslacerda:
I got it!
It works now for every directory that do not contains `__`.

pslacerda:
So lets put more `__` around vars. I failed to use a character unavailable as directory name (eg '\0', `\1`).

orbeckst:
NULL (`\0`) would be good (that's what `find -print0 | xargs -0 ...` uses) – I thought that no-one uses `___` in directory names but I am probably wrong...

Just fix develop, test it locally, and then commit to develop.

Issue #75, opened 2016-07-17
Improve loading times


pslacerda:
Importing GromacsWrapper take seconds. Most of time was wasted in extracting the help for docstrings. Seems that both running the help commands and communicating to acquire the text takes almost 90% of loading times.

This screenshot was taken while [profiling](https://julien.danjou.info/blog/2015/guide-to-python-profiling-cprofile-concrete-case-carbonara) the `gromacs` package importing:

![callgraph](https://cloud.githubusercontent.com/assets/149358/16899430/e98c383a-4bd9-11e6-97fe-cc97083538a5.png)

Albeit it is useful, isn't a such important feature and the library keeps running without it. Or maybe we can load it dynamically and make something similar to this:

```python
class GromacsCommand(Command):
    @property
    def __doc__(self):
        if not self.__doc_cache:
            # load docstring
        return self.__doc_cache
```

Another option would be to reset `GromacsCommand.__doc__` for each subclass. Now we can't simply set `Trjconv.__doc__ = "docstring"` because `Command` is an old style class (eg `class C(object): pass`) which disallow change members of `object` (ie `__doc__`). This change would allow it:

```python
# doesn't inherit from object
class Command:
    pass
```

orbeckst:
I like the idea of loading the doc strings lazily!

pslacerda:
Done =)

Issue #74, opened 2016-07-15
use help(gromacs.tool_name) doesn't works in Gromacs 5


pslacerda:
`gromacs.tool_name.help()` works but `help(gromacs.tool_name)` doesn't

orbeckst:
Funnily enough, in ipython
```
gromacs.tool_name?
```
works and shows the doc string:

```
In [6]: gromacs.grompp?

Type:            Grompp
String form:     <gromacs.tools.Grompp object at 0x105c60910>
File:            /Volumes/Data/oliver/Biop/Projects/Methods/GromacsWrapper/gromacs/tools.py
Signature:       gromacs.grompp(self, *args, **kwargs)
Docstring:
DESCRIPTION

gmx grompp (the gromacs preprocessor) reads a molecular topology file, checks
the validity of the file, expands the topology from a molecular description to
...
```

whereas 
```
In [8]: help(gromacs.grompp)

Help on Grompp in module gromacs.tools object:

class Grompp(gromacs.core.GromacsCommand)
 |  Gromacs tool '%(name)r'.
 |
 |  Method resolution order:
 |      Grompp
 |      gromacs.core.GromacsCommand
 |      gromacs.core.Command
 |      __builtin__.object
 |
 |  Data and other attributes defined here:
 |
 |  command_name = 'grompp'
 |
 |  driver = 'gmx'
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from gromacs.core.GromacsCommand:
 |
 |  __init__(self, *args, **kwargs)
 |      Set up the command with gromacs flags as keyword arguments.
 |
 |      The following  are generic instructions; refer  to the Gromacs
```

pslacerda:
Funny that `getdoc` works but `help` doesn't. Maybe is related to the fact that properties are being used?
```python
import pydoc, gromacs
pydoc.getdoc(gromacs.trjconv)
pydoc.help(gromacs.trjconv)
```

orbeckst:
I looked at  https://bfroehle.com/2012/11/08/property-for-doc-attribute/  (see https://gist.github.com/bfroehle/4041015) which proposes a custom `docstring_property` but when I tried it, it didn't not solve our problem, it just makes a class-level doc string available when there's no generated property doc string.

I also looked on StackOverflow but couldn't find an answer. 

Given that you have a minimum working example do you want to ask on S/O?

pslacerda:
[...]

PR #76, opened 2016-07-17
Load help only on demand

pslacerda:
Hey, just did #75 and now it's loading blazing fast. Was easy after identified where to change!

### DIFF
--- a/gromacs/core.py
+++ b/gromacs/core.py
@@ -383,13 +383,16 @@ class Command(object):
 
 
 class GromacsCommand(Command):
-    """Base class for wrapping a g_* command.
+    """Base class for wrapping a Gromacs tool.
 
     Limitations: User must have sourced ``GMXRC`` so that the python script can
     inherit the environment and find the gromacs programs.
 
     The class doc string is dynamically replaced by the documentation of the
-    gromacs command when an instance is created.
+    gromacs command the first time the doc string is requested. If the tool is
+    not available at the time (i.e., cannot be found on :env:`PATH`) then the
+    generic doc string is shown and an :exc:`OSError` exception is only raised
+    when the user is actually trying to the execute the command.
     """
 
     # TODO: setup the environment from GMXRC (can use env=DICT in Popen/call)
--- a/gromacs/core.py
+++ b/gromacs/core.py
@@ -606,12 +606,16 @@ class GromacsCommand(Command):
         Extract by running the program and chopping the header to keep from
         'DESCRIPTION' onwards.
         """
-        logging.disable(logging.CRITICAL)
+        if self._doc_cache is not None:
+            return self._doc_cache
+
         try:
+            logging.disable(logging.CRITICAL)
             rc,header,docs = self.run('h', stdout=PIPE, stderr=PIPE, use_input=False)
         except:
             logging.critical("Invoking command {} failed when determining its doc string. Proceed with caution".format(self.command_name))
-            return "(No Gromacs documentation available)"
+            self._doc_cache = "(No Gromacs documentation available)"
+            return self._doc_cache
         finally:
             # ALWAYS restore logging...
             logging.disable(logging.NOTSET)
@@ -623,8 +627,11 @@ class GromacsCommand(Command):
             # In GMX 5, the opposite is true (Grrr)
             m = re.match(self.doc_pattern, header, re.DOTALL)
             if m is None:
-                return "(No Gromacs documentation available)"
-        return m.group('DOCS')
+                self._doc_cache = "(No Gromacs documentation available)"
+                return self._doc_cache
+
+        self._doc_cache = m.group('DOCS')
+        return self._doc_cache
 
 
 class PopenWithInput(subprocess.Popen):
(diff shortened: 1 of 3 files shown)


gromacs/core.py
```diff
@@ -495,7 +495,7 @@ def __init__(self, *args, **kwargs):
         self.failuremode = kwargs.pop('failure','raise')
         self.extra_doc = kwargs.pop('doc',None)
         self.gmxargs = self._combineargs(*args, **kwargs)
-        self.__doc__ = self.gmxdoc
+        self._doc_cache = None
```
pslacerda: It was this line that loaded the docstring in the intialization

gromacs/core.py
```diff
@@ -600,44 +600,37 @@ def transform_args(self,*args,**kwargs):
...
-        """Usage for the underlying Gromacs tool (cached)."""
-        if not (hasattr(self, '__doc_cache') and self.__doc_cache):
-            self.__doc_cache = self._get_gmx_docs()
-        docs = self.__doc_cache
-        if self.extra_doc:
-            docs = '\n'.join([self.extra_doc,'',
-                              "Documentation of the gromacs tool", 34*'=',
-                              docs])
-        return docs
+    def __doc__(self):
```
pslacerda: Move to load it only when demanded

gromacs/core.py
```diff
@@ -495,7 +495,7 @@ def __init__(self, *args, **kwargs):
         self.failuremode = kwargs.pop('failure','raise')
         self.extra_doc = kwargs.pop('doc',None)
         self.gmxargs = self._combineargs(*args, **kwargs)
-        self.__doc__ = self.gmxdoc
```
pslacerda: It was this line that extracted the docstring right in the initialization. As a lot of commands classes are created when we import the library, it slows the tool loading.

gromacs/core.py
```diff
@@ -600,44 +600,37 @@ def transform_args(self,*args,**kwargs):
         return self._build_arg_list(**newargs)
 
     def _get_gmx_docs(self):
-        """Extract standard gromacs doc by running the program and chopping the header.
+        """Extract standard gromacs doc
 
-        .. Note::
-
-           The header is on STDOUT and is ignored. The docs are read from STDERR in GMX 4.
```
orbeckst: Leave the note in... it took a while to figure it out and it is valuable to keep this information around. This doc string is not even shown to users so it really is not a problem because it won't confuse them. We could make it a code comment. But in any case, leave it.

orbeckst:
The loading works but now the docs are not available any more (tested with Gromacs 5). Possibly related to #74 

```
import gromacs
In [5]: gromacs.grompp.help()

command: grompp


Gromacs tool '%(name)r'.

In [6]: gromacs.grompp?

Type:           Grompp
String form:    <gromacs.tools.Grompp object at 0x112c92550>
File:           ~/.virtualenvs/gromacswrapper/develop/lib/python2.7/site-packages/gromacs/tools.py
Signature:      gromacs.grompp(self, *args, **kwargs)
Docstring:      Gromacs tool '%(name)r'.
Init docstring:
Set up the command with gromacs flags as keyword arguments.

The following  are generic instructions; refer  to the Gromacs
command  usage information  that should  have  appeared before

In [7]: help(gromacs.grompp)
Help on Grompp in module gromacs.tools object:

class Grompp(gromacs.core.GromacsCommand)
 |  Gromacs tool '%(name)r'.
 |
 |  Method resolution order:
 |      Grompp
 |      gromacs.core.GromacsCommand
 |      gromacs.core.Command
 |      __builtin__.object
 |
 |  Data and other attributes defined here:
 |
 |  command_name = 'grompp'
 |
 |  driver = 'gmx'
 |
 |  ----------------------------------------------------------------------
 |  Methods inherited from gromacs.core.GromacsCommand:
 |
 |  __init__(self, *args, **kwargs)
 |      Set up the command with gromacs flags as keyword arguments.
 |
 |      The following  are generic instructions; refer  to the Gromacs
```

The tool can be run:
```
In [11]: gromacs.grompp(h=True)

                  :-) GROMACS - gmx grompp, VERSION 5.1.2 (-:

                            GROMACS is written by:
```

pslacerda:
This link does exactly the same thing, except that it works on the class itself (eg GromacsCommand) not on the derived one (eg Trjconv).

http://stackoverflow.com/a/2694358

pslacerda:
Sorry!

Please take a look now, it works. As drawback if the user subclass a tool the docstring won't appear. But... someone ever did that?

pslacerda:
And for now it misses only caching

pslacerda:
Oops, no problem in subclasses not inheriting __doc__ as after
```python
class C:
    "Uhaa!"
class B(C):
    pass
```
B.__doc__ is None of course because a derived type hasn't the same doc of its base. So seems that it's working as expected!

gromacs/tools.py
```diff
@@ -128,7 +128,7 @@ def _generate_sphinx_class_string(clsname):
         old_clsname = old_name.replace('.','_').replace('-','_').capitalize()
         cls = type(clsname, (GromacsCommand,), {'command_name': name,
                                                 'driver' :prefix,
-                                                '__doc__': "Gromacs tool '%(prefix) %(name)r'." % vars()})
+                                                '__doc__': property(GromacsCommand._get_gmx_docs)})
```
orbeckst: Very sneaky :-)

orbeckst:
Great work, I am merging this all into one commit.

PR #88, opened 2016-09-12
Added a method to get MDP parameters as dict

whitead:
Wrote a small convenience method to read an MDP file and get a dictionary. 

Also, do we require unit tests yet for PRs? I haven't kept up on status of testing.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,7 @@
 ==============================
 
 2016-xx-xx      0.6.0
-pslacerda, orbeckst
+pslacerda, orbeckst, whitead
 
 * new Gromacs GMXRC parameter in cfg file: if present and set to a
   file then GromacsWrapper sources GMXRC and sets the environment
@@ -12,6 +12,7 @@ pslacerda, orbeckst
   the initial import of the library to almost instantaneously (PR #76)
 * guess which tools to load automatically if no tools option is provided (#68)
 * new documentation page on how to set up a cfg file
+* Can now read MDP files into a dictionary. Useful for joining MDPs. 
 
 2016-06-29      0.5.1
 whitead, dotsdl, orbeckst
--- a/gromacs/cbook.py
+++ b/gromacs/cbook.py
@@ -774,6 +774,36 @@ def get_volume(f):
 # Editing textual input files
 # ---------------------------
 
+def read_mdp(mdp):
+    def mangle(p):
+        """Return a RE string that matches the parameter."""
+        return p.replace('-', '_')  # must catch either - or _
+
+    pattern = re.compile("""\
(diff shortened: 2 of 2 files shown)


whitead:
Just realized that there is an MDP file format class that can do what I need. Maybe this is redundant...

orbeckst:
Did you check `fileformats.mdp.MDP`?  Does it do what you need or is it worthwhile merging your approach with `MDP`?

Finally, @pslacerda also has a pretty sleak MDP data structure in his [gmxscript](https://github.com/pslacerda/gmx) and I vaguely remember that at some point he was talking about porting it to GW.

Regarding tests: I am now trying to add more tests and certainly tests for new features are *very much encouraged* (and tests for old stuff will net you lots of good karma ;-) ).

whitead:
Yeah, I realize now that the `MDP` class does everything I need. When I have some free time, I'll put a reference in the `cbook.edit_mdp` doc to the `MDP` fileformat. And maybe write some of these karma generating tests :+1:

pslacerda:
What can a user want to do with a mdp file? The [sample.py][1] shows only part of what we can put on GromacsWrapper.

We can create something like the following hypothetical `mdp()` function: 

```python
# reads the file step.mdp
>>> mdp('step.mdp')
{'integrator': 'steep', 'emtol': 10}

# returns the path of a temporary file with parameters from step.mdp
# merged with user supplied ones.
>>> mdp('step.mdp', {
...  'rlist': 1.4,
...  'rcoulomb': 1.4,
...  'rvdw': 1.4
...  })
{'integrator': 'steep', 'emtol': 10, 'rlist': 1.4, 'rcoulomb': 1.4, 'rvdw': 1.4}

# returns the path of a temporary file created with these parameters
>>> mdp({'integrator': 'steep', 'emtol': 10, 'rlist': 1.4, 'rcoulomb': 1.4, 'rvdw': 1.4})
{'integrator': 'steep', 'emtol': 10, 'rlist': 1.4, 'rcoulomb': 1.4, 'rvdw': 1.4}

```

Okay, maybe three separated functions.

[class MDPResource][1] also features automatic lookup in parent directories: `mdp('mdp/step.mdp')` returns `mdp/step.mdp`, `../mdp/step.mdp`, `../../mdp/step.mdp`, or raise an exception if not found.

Tests are very very nice, but also very hard. There are some crazy people that likes it so much that tests before developing (ie TDD), they can think how the new artifact will look beforehand.

Hope we can achieve a neat and pleasurable tool for molecular dynamics.

[1]: https://github.com/pslacerda/gmx/blob/master/sample/sample.py
[2]: https://github.com/pslacerda/gmx/blob/master/gmxscript.py#L193

Issue #44, opened 2016-04-05
support Python 3


whitead:
Are you planning on adding Python 3 support? I see some mention of 2to3 in the commits, not couldn't find a description anywhere else. Thanks!

dotsdl:
I think we'd like to, but the library doesn't have regular development at the moment. @orbeckst can we get a rough list of what needs to get done to make GromacsWrapper work under Py3? Then we can get to work knocking these components out piecemeal.

orbeckst:
Not sure what is needed at the moment; it;s probably not a big deal because there's no C/Cython code. Using six will probably do the job. 

Perhaps run 2to3 and check if that's enough to get it to import, then use six to make it work under 2 and 3.


On 6 Apr, 2016, at 10:42, David Dotson wrote:

> I think we'd like to, but the library doesn't have regular development at the moment. @orbeckst can we get a rough list of what needs to get done to make GromacsWrapper work under Py3? Then we can get to work knocking these components out piecemeal.
> 
> �
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> 

--
Oliver Beckstein * orbeckst@gmx.net
skype: orbeckst  * orbeckst@gmail.com

orbeckst:
I had a look and from browsing the `2to3` patches it seems that we can do pretty much everything with a `six` compatibility layer. It's a bit of boring work but doable.

dotsdl:
I have good motivation to knock this one out with [`alchemlyb`](https://github.com/alchemistry/alchemlyb) development. I'll make a python3 branch so I can get tests via travis easily.

orbeckst:
@dotsdl the develop branch has been cleaned up (gromacs.analysis gone, numkit separate). If you want to have a go at Py 3 then please do! Note that numkit also needs to become Py3, see Becksteinlab/numkit#1.

orbeckst:
@frchalaoux please open a *pull request* that includes your patches from #121.

If you haven't done this on GitHub yet:
1. fork the project to your own GitHub account
2. make a branch in your fork (e.g., issue44-py3)
3. create a pull request from your repo against the GromacsWrapper *develop* branch

We'll then start reviewing and discussing the code. The whole review process will happen on your PR.

frchalaoux:
@orbeckst 
I would like to ask you for advice to decide if my starting architecture is practitable ? : 
- A fork of your GromacsWrapper on my Github (my repo Github)
[...]

Issue #82, opened 2016-08-03
Split the plugins into a separated repository


pslacerda:
We can move the plugins to a separated repository, it will be cleaner and make easier to move just the main code to Python 3.

pslacerda:
Based on [namespace packages]][1] we can make the plugins into a separated repository keeping it at the same `gromacs` namespace.

1. Remove the `gromacs/analysis/plugins/` directory
2. Delete the line `import plugins` from `gromacs/analysis/__init__.py`
3. Install the whole package
4. Install the [plugins pakcage][2]

Now we can

```python
import gromacs.plugins
```

from a separated repository =). Or is better keep the `gromacs.analysis.plugins` namespace because all plugins are for analysis?

[1]: http://morepath.readthedocs.io/en/latest/organizing_your_project.html#namespace-packages
[2]: https://github.com/pslacerda/GromacsWrapper-plugins

pslacerda:
The main importance of plugins are to enable parallel analysis right? I heard about some guys that parallelized frame-by-frame analysis splitting the trajectories and submitting a job for each part then combining the results. They used Spark to do it on multiple computers but I don't go that far:

```sh
gmx rmsf -b    0 -e 1000 -o rmsf_0 &
gmx rmsf -b 1001 -e 2000 -o rmsf_1 &
```

With this very simple trick is possible to enable parallel analysis. The operating system takes care to allocate the resources intelligently. In most cases combining the results is just a simple concatenation as in .xvg files above. If I remember correctly from the lists the Gromacs team is also pursuing trivial analysis parallelization like this by default.

If we do:
```python
def figure_out_length(f):
    return 1000

def parallel_analysis(tool, njobs, **kwargs):
    begin = kwargs.get('b', 0)
    end = kwargs.get('e', None)
    if end is None:
        end = figure_out_length(kwargs['f'])
    
    kwargs_list = []
    count = 0
    for part_begin in range(begin, end, (end-begin)//njobs):
        part_end = part_begin + ((end-begin)//njobs) - 1
        if part_end > end:
            part_end = end
        part_kwargs = kwargs.copy()
        part_kwargs['b'] = part_begin
        part_kwargs['e'] = part_end
        for key, value in kwargs.items():
            if isinstance(value, str) and '%d' in value:
                part_kwargs[key] = value % count
        kwargs_list.append(part_kwargs)
        count = count + 1
    return kwargs_list
```

And then:
```python
from gromacs import *
[...]

PR #101, opened 2016-12-29
WIP: Issue #44 Python 3

dotsdl:
We want this library to work under both Python 3 and 2 (issue #44). I've started by running `2to3` on the whole `gromacs` module, and will troubleshoot each issue using feedback from the tests via Travis. Will remove WIP tag when finished.


dotsdl:
We now have builds of gromacs conda packages for both gromacs 4.6.7 and gromacs 5.1.2 for python 3.5, but the thing that fails now is getting `recsql`, which [also doesn't work for python 3 yet](https://github.com/orbeckst/RecSQL/issues/5). I think we have at least two options:
1. Solve orbeckst/RecSQL#5 
2. Remove numkit (#82), which I believe is the only thing that really needs `recsql`?

What do you think we should do, @orbeckst?

orbeckst:
Preferrably 1 and fix orbeckst/RecSQL#5... but the problem is not numkit, I think (where did you find recsql there??) but more likely gromacs.analysis.... and that we would like to split off in #82.

So I think from the GW perspective: solve #82 and this one solves itself...

orbeckst:
We want to remain compatible with Python 2.7 for right now. Therefore, simple `2to3` is not good, for instance, we need to keep the `from __future__ import...`.  Perhaps `2to3` has options to remain compatible?

The testing matrix is good.

orbeckst:
PR is superceded by PR #127

PR #102, opened 2017-01-27
Gromacs 2018 and pytest low-performance custom plugin to control -nt

### LINKED ISSUE #96: test Gromacs 2018
We need to test _Gromacs 2018_.
- [x] Gromacs 2018 conda package (PR #127 )
- [x] add tests to travis matrix
- [x] fix tests... possibly raise new issues

EDIT: use 2016 instead of 2018

ianmkenney:
~Include Gromacs 2016 in tests (fixes #96)~

EDIT: We use Gromacs 2018 (see PR #127 and discussion in #130 ). This PR contains experimental pytest plugin to provide the `pytest --low-performance` flag and code to explicitly set the number of threads for `gmx mdrun` in the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,9 +49,6 @@ matrix:
 install:
   - git clone --depth 1 git://github.com/astropy/ci-helpers.git
   - source ci-helpers/travis/setup_conda.sh
-  # install test tools (explicit instead of PIP_DEPENDENCIES because
-  # this is temporary until #137 is solved)
-  - pip install git+git://github.com/ianmkenney/pytest-gmx.git  
   - eval $BUILD_CMD
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ env:
 matrix:
   fast_finish: true
   allow_failures:
-    - env: GROMACS_VERSION=2018.2  PYTHON_VERSION=2.7 CODECOV=false
     - env: GROMACS_VERSION=2018.2  PYTHON_VERSION=3.6 CODECOV=false
     - env: GROMACS_VERSION=4.6.5   PYTHON_VERSION=3.6
     - os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   global:
     - PYTHON_VERSION=2.7
     - CODECOV=true
-    - PYTEST_FLAGS="-v --disable-pytest-warnings --durations=20 --low-performance"
+    - PYTEST_FLAGS="-v --disable-pytest-warnings --durations=20 --low-performance --cov=gromacs"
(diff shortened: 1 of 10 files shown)


orbeckst:
Similar problem with maxthreads as in Becksteinlab/MDPOW#77 ?

orbeckst:
Can you enable full output so that we can see if this is again a problem with the number of threads? We probably need to explicitly restrict number of threads to 2 again...

orbeckst:
@ianmkenney can you have a look again, please? 

I would like this to work.

orbeckst:
What is pytest-gmx and where is it used?

ianmkenney:
I'm attempting to build a plugin for pytest that will make our tests a little more manageable. Setting the number of threads from the command line rather than hard coded into the tests themselves would be much more convenient since when we run local tests we would rather use a larger number of threads. 
For now I just added a low performance flag to pytest that, if provided, will set nt=2. Just an playing around with the idea, as it could still be useful in future projects. You can find it in my repos.

orbeckst:
@ianmkenney can you move the repo to Becksteinlab so that all the GW-related things are together?

orbeckst:
Rebased against develop but there are problems in the tests
- [2016 test](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/372286811) times out
- [5.1.2 test](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/372286810#L831) (and the 4.6.7 test) have an error in `test_mdrun()` with the wrong number of arguments.

orbeckst:
rebased against develop

orbeckst:
Btw, locally, the tests pass just fine with 2016.5.

orbeckst:
Good and bad news:
* Good: for the first time, *all Gromacs 2016* tests run (probably due to using `--low-performance`, i.e., setting `gmx mdrun -nt 2` instead of letting Gromacs choose the number of threads).
* Bad: tests still fail
  * [Gromacs 2016 tests](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/373712379) (However, my local version of Gromacs 2016.5 passes the tests just fine.)
      * pdb2gmx generates a system with a non-integer charge (?!?!?!?!) – maybe check the occupancy fields of the PDB file. pdb2gmx complains about a few of them. 
      * cg minimization fails with a constraint problem
  * [Gromacs 4.6.7 tests](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/373712377): tests that used to pass, fail now:
       * cg energy minimization fails (not sure why)

orbeckst:
This looks good, and the run time of the mdrun tests decreased from ~20s to ~4s by setting -nt 2. Will merge.

PR #103, opened 2017-01-30
Lists for MDP files

PicoCentauri:
Hey, I added some lines to convert lists in MDP files to bumpy arrays. Hopefully this is the correct branch...


orbeckst:
@PicoCentauri thanks for the patch. Can you please 
- describe if the patch fixes a bug (if so, how does one produce the bug) or if it is an enhancement (and why is the enhancement needed)
- add test cases that test the new functionality
   - L72 in  tests/test_utilities: [test_autoconvert](https://github.com/Becksteinlab/GromacsWrapper/blob/develop/gromacs/tests/test_utilities.py#L72)
   - add a test for MDP reading ... I'll quickly add a skeleton test to tests/fileformats

orbeckst:
I don't think that you are on the correct branch. You should use the *develop* branch for your work instead of *master*. Your PR should only contain your own commits. Perhaps just `git rebase` your branch against develop (if you know how to do this – otherwise I would checkout develop and just `git cherrypick 8e14016` onto your develop branch.

PR #122, opened 2017-05-08
First shot from FRC to permit GW a python 2 and 3 compatibility: d…

frchalaoux:
…el statement was not permit in python 3 if the object does not exists. A try should be sufficient.

### DIFF
--- a/gromacs/__init__.py
+++ b/gromacs/__init__.py
@@ -278,7 +278,12 @@ if len(_missing_g_commands) > 0:
                   "maybe source GMXRC first? The following are missing:\n%r\n" % _missing_g_commands,
                   category=GromacsImportWarning)
 
-del name, cls, clsname
+#Python 2 and 3 compliant
+try:
+	del name,cls, clsname
+except NameError:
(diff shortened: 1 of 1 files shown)


frchalaoux:
>> Merging #122 into develop will decrease coverage by <.01%. What does it mean ? Sorry to be so lazy.

frchalaoux:
It was only a try. Thanks to all. 
FR.

orbeckst:
https://github.com/Becksteinlab/GromacsWrapper/pull/122#issuecomment-299993412 : It basically means that testing coverage decreased but 0.01% is nothing. In general we would like for test coverage to go *up*, which typically means committing tests together with code changes. 

But don't worry about the coverage yet.

orbeckst:
Also note that @dotsdl has tried a bit of Py3 compatibility in PR #101 – have a look at it. I don't think we are going to try and merge #101 because too much changed. But I might lift a few things into a new PR. You can certainly have a look at what @dotsdl already did to get an idea of what needs to be done.

PR #127, opened 2018-04-26
Python3

### LINKED ISSUE #44: support Python 3

kain88-de:
except for two unit tests everything passes.  Most changes should be harmless except for the tool registry, could be I broke that. But changing a dict during iteration isn't allowed in python 3 anymore.

fixes #44


orbeckst:
The Python 2.7 tests fail because of
- `six.long` does not seem to be found
- `DataFrame` does not have `sort()` (?!?)

The Python 3 tests have problems installing Gromacs. I am looking into it.

gromacs/cbook.py
```diff
@@ -974,7 +975,7 @@ def edit_txt(filename, substitutions, newname=None):
             else:
                 logger.debug("Deleting line %r", line)
 
-    target.seek(0L)
+    target.seek(six.long(0))
```
orbeckst: `six.long()` does not exist in six

gromacs/cbook.py
```diff
@@ -1040,7 +1041,7 @@ def remove_molecules_from_topology(filename, **kwargs):
                 continue  # remove by skipping
             target.write(line)
     if autogenerated and n_removed > 0:
-        target.seek(0L)
+        target.seek(six.long(0))
```
orbeckst: `six.long()` does not exist in six

gromacs/run.py
```diff
@@ -332,7 +333,7 @@ def check_mdrun_success(logfile):
     except IOError:
         return None
     try:
-        log.seek(-1024L, 2)
+        log.seek(six.long(-1024), 2)
```
orbeckst: `six.long()` does not exist in six

gromacs/cbook.py
orbeckst:
```
In [4]: six.long
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-daab08199c2e> in <module>()
----> 1 six.long

AttributeError: 'module' object has no attribute 'long'

In [5]: six.__version__
Out[5]: '1.11.0'
```

orbeckst: I don't think we actually need the longs... so I am removing them.

kain88-de:
If it works without the longs I would not use them anymore. Looks like the challenge now is python3 for your gromacs packages.  I'm not sure what you mean by the DataFrame.sort issue.

orbeckst:
Some tests used DataFrame.sort(), which does not exist anymore. But I fixed that. 

As you said, the Gromacs conda packages are the issue. The Gromacs devs are interested in creating their own conda packages but until that happens I still have to do something. When we hacked together our Gromacs conda packages we did not know about conda-forge but we should probably have invested the time to learn about it---it would be super-convenient if conda-forge could build the packages. 

--
Oliver Beckstein
email: orbeckst@gmail.com

kain88-de:
> Some tests used DataFrame.sort(), which does not exist anymore. But I fixed that.

oh yeah this was replaced by `sort_values`. I was looking for this last week myself.

about the gromacs packages. Are there some emails about this on the mailing list? We do have some good contacts conda-forge to help out with this.

orbeckst:
> about the gromacs packages. Are there some emails about this on the mailing list? We do have some good contacts conda-forge to help out with this.

No, that wasn't a public discussion. I can email you offline if you like.

kain88-de:
you can email me offline thanks.

kain88-de:
yay so python3 is running for at least one gromacs version

kain88-de:
Ok so with the new packages from bioconda the tests seem to run. There are still failures but there is progress.

orbeckst:
@ianmkenney has the same problem with Gromacs 2018 on conda in PR #130.

kain88-de:
most things are done now. Some string handling issues are left.

kain88-de:
@orbeckst this now sets all failing jobs to be allowed to fail. The old tests still work. Would you like to merge this as is? The last issues are with string handling but that might be better handled by someone who has used GromacsWrapper more then I did (I only needed the xvg reader once)

orbeckst:
I'd like to get the Python 3/Gromacs 4.6.5 tests to pass so that this PR really fixes #44.

I'll try to look at it over the next few days.

kain88-de:
I'm down to three errors now. The last ones fail when gromacs is rerun for amber forcefield. So something internally is messed up. Fixing these errors is not a lot more difficult then pocking in the debugger and some light googling.

kain88-de:
ok so yeah string handling fixes for python3 have broken python2 -.-

kain88-de:
@ianmkenney do you know why the last three tests fail?

orbeckst:
I have no idea why these [Test<FF>.test_mdrun  tests fail under 3.6](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/380659601#L823) but pass under 2.7. Why are the numerical results different? However, this looks as if it could be an easy win if it's something obviously stupid. ... and then this PR can be merged and we can start fixing 2018.

kain88-de:
I would need some help to investigate those errors. It would be helpful to know the exact gromacs commands that are executed for this test in python 2.7 and 3.6. Without knowing the library it's not straightforward to understand the program flow that is happening.

kain88-de:
@ianmkenney did you have a chance to fixup the gromacs errors?

kain88-de:
I don't mind if you want to merge this. Just keep in mind I don't have much
time nor expertise to fix things in this project.

orbeckst:
I totally understand.

orbeckst:
I am rewriting the history to clean up the commits. I will force-push to your branch and then merge.

PR #130, opened 2018-05-09
Changed gromacs package for testing

ianmkenney:
Moved from becksteinlab to bioconda package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ branches:
         - develop
 
 env:
-    - GMX_VERSION=4.6.7
-    - GMX_VERSION=5.1.2
+    - GMX_VERSION=4.6.5
+    - GMX_VERSION=2018
 
 # install python dependencies
 install:
@@ -36,8 +36,13 @@ install:
   - conda create --yes -q -n gw python=$TRAVIS_PYTHON_VERSION numpy scipy pandas
   - source activate gw
 
-  # install becksteinlab Gromacs build
-  - conda install -c becksteinlab gromacs-$GMX_VERSION
+  # install bioconda Gromacs build... Need to set up channels first
+  # per instructions at: https://bioconda.github.io/
+
+  - conda config --add channels defaults
+  - conda config --add channels conda-forge
+  - conda config --add channels bioconda
+  - conda install gromacs=$GMX_VERSION
 
   # install other dependencies through pip
   - pip install numkit
@@ -48,11 +53,9 @@ install:
   - pip install pytest-pep8
   - pip install -e .
 
-  # source gromacs
-  - source $($HOME/get_gmx $GMX_VERSION)
-
 # run tests
 script:
+  - gmx grompp --help
   - py.test --cov gromacs gromacs/tests
 
 after_success:
--- a/environment.yml
+++ b/environment.yml
@@ -1,8 +1,8 @@
 name: gw
 channels:
-- becksteinlab
+- bioconda
 dependencies:
-- gromacs-5.1.2
+- gromacs
 - numpy>=1.9
 - scipy
 - pandas


.travis.yml
```diff
@@ -36,8 +36,13 @@ install:
   - conda create --yes -q -n gw python=$TRAVIS_PYTHON_VERSION numpy scipy pandas
   - source activate gw
 
-  # install becksteinlab Gromacs build
-  - conda install -c becksteinlab gromacs-$GMX_VERSION
+  # install bioconda Gromacs build... Need to set up channels first
+  # per instructions at: https://bioconda.github.io/
+
+  - conda config --add channels defaults
```
kain88-de: defaults is always on by default.

environment.yml
```diff
@@ -1,8 +1,8 @@
 name: gw
 channels:
-- becksteinlab
+- bioconda
```
kain88-de: you also need to add conda-forge for the version 2018 dependencies

kain88-de:
maybe continuing to test gromacs 5.x from the beckstein lab repo would be good. there is still a large number of people using gromacs 5

orbeckst:
The problem with the 5.x build is that it is currently tied to Python 2; we would need to build a Python 3 version (or get rid of any Python dependencies - which is the `get_gmx` script). We hope to not have to maintain the gromacs conda packages any more...

orbeckst:
@ianmkenney said that the Bioconda gromacs 2018 core dumps on Travis – perhaps an issue with the instruction set not matching what is available on Travis??

@ianmkenney can you post the evidence that you have?

kain88-de:
I can imagine where the core dumps come from. The travis/circleci build on conda forge was done with a random CPU and GROMACS always chooses the best available SIMD instruction set. On travis we can be randomly assigned an older CPU that doesn't support the instruction set. The only way around this is to specify a common instruction set like SSE3 during the cmake step. Making our own package on conda-forge can help to sort out these issues.

About maintenance. Using bioconda and conda-forge doesn't get rid of it completely but sharing responsibility is much easier.

ianmkenney:
[log with the error](https://travis-ci.org/Becksteinlab/GromacsWrapper/jobs/377026703)

I put `gmx grompp --help` and got an illegal instruction error

```
$ gmx grompp --help
/home/travis/.travis/job_stages: line 57:  2239 Illegal instruction     (core dumped) gmx grompp --help
```

kain88-de:
yeah that is a instruction issue. But that is fixable upstream

orbeckst:
WHat is the best approach to proceed? Raise an issue with bioconda and ask them to build Gromacs taking [portability](http://manual.gromacs.org/documentation/2018/install-guide/index.html#portability-aspects) into account? E.g. build binaries with instruction set suffixes
- `gmx_sse2`
- `gmx_avx`
- etc
or just build `gmx` with sse2?

kain88-de:
I'm submitted a small PR.

kain88-de:
I would have gone for SSE3 but gromacs doesn't over it.  I'm aware of the conda-build3 transition happening in conda-forge. It will still take a little while for them to make it though.

kyleabeauchamp:
SSE2 SGTM then

kain88-de:
@ianmkenney you should be able to try in a about an hour.

rhheilma:
Can I get off of this list, please?

Thank you,
Ryan

kain88-de:
@rhheilma you have to change your GitHub subscription setting yourself. We cannot do this for you.

orbeckst:
The Python 3 PR #127 includes the new bioconda packages so we can fix everything there (and close this PR).

Thanks @kain88-de for fixing upstream.

PR #134, opened 2018-08-02
properly handle bytes output from subprocess in auto tool detection

### LINKED ISSUE #133: tools not loaded for Gromacs 2016/2018 in Python 3 on macOS
When using Gromacs 2016/2018 (the `gmx` command) and Python 3 (and apparently only on macOS) then the tools are not correctly loaded; only `tools.Gmx` appears.

Reason: byte string snafu

```python
            out = subprocess.check_output([driver, '-quiet', 'help',
                                           'commands'])
            for line in str(out).encode('ascii').splitlines()[5:-1]:
```
The `subprocess.check_output()` returns a `b` string, 
```python
>>> type(out)
bytes
```
which, when converted with `str()`, does not maintain the newlines but rather includes them quoted
and then `splitlines()` does not work anymore.

orbeckst:
- fix #133
- affected macOS with Python 3
- added test

### DIFF
--- a/gromacs/config.py
+++ b/gromacs/config.py
@@ -648,7 +648,7 @@ def set_gmxrc_environment(gmxrc):
         out = subprocess.check_output(cmdargs)
         out = out.strip().split()
         for key, value in zip(envvars, out):
-            value = value.replace('___', '')  # remove sentinels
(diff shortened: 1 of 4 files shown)


orbeckst:
There are still various failures unrelated to this PR so I will merge it.

PR #144, opened 2018-08-09
try caching miniconda and pip on Travis CI

orbeckst:


orbeckst:
Does not work yet, probably needs to be solved in https://github.com/astropy/ci-helpers/issues/207

PR #164, opened 2018-09-06
 Added function for setting up FEP simulations

PicoCentauri:
The function in the setup module allows to set up free energy pertubation simualtions. It will create a single simulation folder for all lambda states to run them individually.

In contrast to the other setup functions I used a dictionary input for the submission script, since I think its more convenient. 

Furthermore I wondered about two thinks: 

1. The setup functions seem not really consistent regarding their input parameters and their standard values between them: for example [energy_miniztion](https://github.com/Becksteinlab/GromacsWrapper/blob/d96c34f5d7d81175e823746322f2760c31447e81/gromacs/setup.py#L609-L655) and [MD](https://github.com/Becksteinlab/GromacsWrapper/blob/d96c34f5d7d81175e823746322f2760c31447e81/gromacs/setup.py#L1034-L1087). Maybe we should think about to unify those. 

2. Is there a reason that the [edit_mdp](https://github.com/Becksteinlab/GromacsWrapper/blob/d96c34f5d7d81175e823746322f2760c31447e81/gromacs/cbook.py#L778-L899) function is not using the mdp file format parser?

### DIFF
--- a/gromacs/setup.py
+++ b/gromacs/setup.py
@@ -35,7 +35,7 @@ sequence of functions that depend on the previous step(s):
         :func:`energy_minimize`)
   :func:`MD_restrained`
         set up restrained MD
-:func:`MD_restrained`
+  :func:`MD_restrained`
       set up free energy pertubation MD
   :func:`MD`
         set up equilibrium MD
@@ -1046,11 +1046,11 @@ def MD_fep(dirname='MD_FEP', stateprefix="lambda_", **kwargs):
 
     :Keywords:
        *dirname*
-          set up under directory dirname [MD_POSRES]
+          set up under directory dirname [MD_FEP]
        *stateprefix*
-          Folder prefix for the lambda state
+          Folder prefix for the lambda state [lambda_]
        *struct*
-          input structure (gro, pdb, ...) [em/em.pdb]
+          input structure (gro, pdb, ...) [conf.gro]
        *top*
           topology file [top/system.top]
        *mdp*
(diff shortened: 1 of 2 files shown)


orbeckst:
> Is there a reason that the edit_mdp function is not using the mdp file format parser?

No, just history... the parser came later and I never got around to rewrite something that seemed to work even it was crappy code.

PicoCentauri:
> No, just history... the parser came later and I never got around to rewrite something that seemed to work even it was crappy code. 

Okay, I can take a look to decrease "duplicated" code and reduce probable sources of errors.

orbeckst:
@PicoCentauri sorry to revive an old PR after such a long time – if you're still interested in getting this function in then the main thing would be to have some sort of tests. That's not that easy but would really be good.

Btw, I would very much encourage using the MDP parser instead of of `edit_mdp()`...

gromacs/setup.py
```diff
@@ -1031,6 +1035,93 @@ def MD_restrained(dirname='MD_POSRES', **kwargs):
...
+            
+            energy_minimize(dirname=dirname_state, mdp=em_mdp,
+                            struct=struct_state, top=top, 
+                            output='em.gro', deffnm="em",
+                            init_lambda_state=l, **kwargs)
+            struct_state = os.path.join(dirname_state,"em.gro")
+        
+        new_kwargs = _setup_MD(dirname=dirname_state, mdp=mdp, struct=struct_state, 
+                               top=top, ndx=ndx, mainselection=mainselection, 
+                               init_lambda_state=l, **qsub_opts, **kwargs)
```
orbeckst: Double `**kw1, **kw2` does not work in Python 2.7 – needs to be rewritten.

PicoCentauri:
@orbeckst Never mind for reviving this PR. In principle yes I'm still willing to merge this PR. I will take a look at the issues and the tests this week.

PicoCentauri:
Sorry currently I have no time to keep this PR a live and add sufficient tests.

PR #158, opened 2018-08-18
JOSS paper

orbeckst:
Short software paper for submission to the [Journal of Open Source Software](http://joss.theoj.org/).

The paper will be part of the GromacsWrapper repository; this appears to be the common pattern with other JOSS papers.

- [submission instructions](https://joss.readthedocs.io/en/latest/submitting.html)
- [review criteria](https://joss.readthedocs.io/en/latest/review_criteria.html)

Closes Issue #147 (This branch will be merged when the paper is submitted.)

# Author contributions
- read #147 for criteria for inclusion
- reply to #147 and state that you want to be an author
- for right now: fork the repo and submit PRs against the *paper* branch or try to edit the `paper.md` directly on GitHub.

Issue #147, opened 2018-08-09
JOSS paper


orbeckst:
Write a short software paper on GromacsWrapper for [JOSS](http://joss.theoj.org/).
- [Author instructions](https://joss.readthedocs.io/en/latest/submitting.html) – create a repo and then submit, ~1000 words, see examples
- authors? 
  - any contributor (name on a commit)
  - must edit and ok paper

richardjgowers:
I'd be interested in this.  I'm going to try and get coverage higher (and find any easy bugs)

orbeckst:
The more coverage the more Unicode issues raise their head....

So yes, more tests would be super!

--
Oliver Beckstein
email: orbeckst@gmail.com

orbeckst:
@pslacerda @jandom @kain88-de @dotsdl @whitead @richardjgowers  @ianmkenney @AndySomogyi  @PicoCentauri @mohebifar – thank you for contributing to GromacsWrapper. I'd like to write a short [Journal of Open Source Software](http://joss.theoj.org/) paper. *If you are interested in being an author on this mini-paper then please reply on this issue.* 

At a minimum, you will need to read the draft of the paper, make corrections if necessary, and ok it. I would also expect willingness in helping to respond to referees' requests for revisions.

I will create a repo for the paper and invite everyone who wants to be an author. Anyone who does not want to be an author will be acknowledged in the acknowledgements (unless they do not want to be mentioned at all). 

The top of this issue has links to JOSS and the author instructions.

— @orbeckst

Oliver Beckstein, DPhil * @orbeckst
https://becksteinlab.physics.asu.edu/

Associate Professor of Physics
Arizona State University
Center for Biological Physics and Department of Physics
Tempe, AZ 85287-1504
USA

Department of Physics: https://physics.asu.edu/content/oliver-beckstein
Center for Biological Physics: https://cbp.asu.edu/content/oliver-beckstein

AndySomogyi:
Hi Oliver.

I not sure I’ve contributed a huge amount, but I’d very much like to help in writing and editing a paper. Let me know what you’d like me to do.

PicoCentauri:
Following @AndySomogyi I would also be happy if I can contribute anything to the paper!

pslacerda:
Of course I want!

I'm had almost forgot about MD but in the last months I did some modelling and I very want to do more (and submit the JOSS paper).

[...]

PR #165, opened 2019-02-10
Add option to ignore suffix of gmx executables

### LINKED ISSUE #124: Using differently named gromacs commands
My installation of gromacs has commands named, for example, `gmx_mpi` (all ending `_mpi`). These then get used in this tools as, for example, `gromacs.grompp_mpi`. This normally works fine for me, but when trying to import `gromacs.setup`, there are two commands that are issues: one on line 585 of  `cbook.py` and one on line 163 of `setup.py`. One calls `tools.Grompp` (which for me needed to be `tools.Grompp_mpi`) and the other calls `gromacs.tools.Trjconv` (which needed to be `gromacs.tools.Trjconv_mpi`).

I'm not sure if there is a way to automatically find what these should be called based on how the functions are defined, or if this is just something I need to deal with (and manually fix) because of my differently-named gromacs commands.

theavey:
In response to my very old issue, this fixes #124.

This adds an option in the config file called append_suffix. If it is set to 'no', the suffix of the gmx executables (such as "_mpi" or "_d") will be ignored when making the tool names. It's not ideal to not differentiate them, but it can make code much more consistent across computers.

The largest change here is that in order to properly test this, the config file has to be changed before gromacs is imported. pytest makes all imports relative to the highest package it can find. Therefore, when `tests` was nested in the `gromacs` folder, the tests would be imported as `gromacs.tests.test_...`. This means that the `gromacs.__init__.py` is run before even running any of the pytest setup functions. By moving the tests folder out of the gromacs package, the config file can be edited before the `gromacs.__init__.py` is ever called.

Because of this change, I've tried to refactor all the references to `gromacs.tests` or `gromacs/tests` that I could find. 

This PR another item to the travis testing matrix that tests this option. Using two new command line options to pytest, a link to the gmx executable called 'gmx_mpi' is put into the home folder and written in the tools section of the config file; and the config file is written to ignore the suffix of the gmx executables found.


theavey:
It seems that I was overly ambitious trying to make what I thought were trivial changes to the ConfigParser and some uses of it.

Seeing as you seem to want to continue Python 2 support, I can roll those changes back.

tests/test_config.py
```diff
@@ -14,7 +14,13 @@
 import gromacs.config
 import gromacs.utilities
 
-def get_GMXRC():
+
+@pytest.fixture
```
theavey: Made GMXRC a fixture

tests/test_config.py
```diff
@@ -26,7 +32,7 @@ def get_GMXRC():
     bindir = os.path.dirname(path)
     GMXRC = os.path.join(bindir, 'GMXRC')
     if not os.path.exists(GMXRC):
-        raise IOError(errno.EEXIST, "Could not find Gromacs setup file", GMXRC)
+        raise IOError(errno.ENOENT, "Could not find Gromacs setup file", GMXRC)
```
theavey: This now raises the error that the file does not exist instead of that it does exist (it would be returned if it did exist).

richardjgowers:
@theavey what's the problem you're running into with Py2?  I took a quick look and it looks like the `super()` call isn't happy, but I don't quite get why, it looks fine to me

theavey:
So in Python 2, ConfigParser (or SafeConfigParser) are old-style classes (they don't inherit from `object`), so they can't be imported that way. If they were new-style classes, that syntax is correct.

It seems that one option is to add a second parent of `object`. I've done this and a couple other fixes for Python 2 in a different branch. If the tests pass, I'll pull those changes into my main branch so they will be reflected here.

theavey:
If the reason for moving the tests folder wasn't clear, I just wrote up a [stackoverflow answer](https://stackoverflow.com/a/54640587/3961920) on a very related topic that might explain it more clearly.

theavey:
Hi @orbeckst, this PR is not critical, but if any changes need to be made, sooner would be better for me. This is just fresh in my mind now, and it would be easier to work on now instead of having to get back into this later. Thanks!

.travis.yml
```diff
@@ -51,6 +54,9 @@ install:
 script:
   - cd ${TRAVIS_BUILD_DIR}
   - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then ulimit -S -n 2048; fi
+  - if [[ $SUFFIX_TEST == 'true' ]]; then
```
richardjgowers: this if is messy because the extra test config is defined in 3? different places across the yaml.  Could it work so that the extended `${setup_cmd}` is defined in the `matrix:` section only?  Ideally something like `setup_cmd = ${setup_cmd} + '--no-append-suffix'`

gromacs/config.py
```diff
@@ -234,9 +234,22 @@
...
 
 if sys.version_info[0] < 3:
-    from ConfigParser import SafeConfigParser
+    from ConfigParser import SafeConfigParser as ConfigParser
+    from ConfigParser import NoSectionError, NoOptionError
+    ConfigParser.read_file = ConfigParser.readfp
+    _cf_getbool = ConfigParser.getboolean
+    _UNSET = object()
+
+    def _getboolean(self, section, option, fallback=_UNSET, **kwargs):
```
richardjgowers: this deserves a comment or two..

.travis.yml
theavey: Makes sense. I'll clean it up.

orbeckst:
@theavey sorry, I hadn't seen the PR – just very busy at the moment, so wasn't checking my GitHub feed. Generally speaking, if you need a reaction include @orbeckst so that I get an email notification.

@richardjgowers is already looking at the PR (THANK YOU Richard!!) and I won't really have time until the end of the week so whatever @richardjgowers decides is fine with me.

orbeckst:
And yes, I'd like to keep Python 2.7 compatibility – too much old code that uses it.

tests/test_config.py
orbeckst: oops. thanks for catching

orbeckst:
@theavey sorry for the slow responses from me. Could you please resolve the conflicts that are flagged here?

orbeckst:
Please ping me or @richardjgowers when we should review again. Thanks!

theavey:
Hi @orbeckst,

The conflicts were just that I had moved the tests.

I've added more comments on `ConfigParser.getboolean`. The main change here is that I backported the `fallback` kwarg for Python 2.

The mode changes were inadvertent. I think they must have just been 644 when I cloned the repository. I am not sure how else they would have been changed.

I've added to CHANGES and AUTHORS.

orbeckst:
@theavey many thanks for your contribution!

theavey:
Of course! Thank you for the package

orbeckst:
P.S.: We wanted to write a short JOSS paper #147 – if you're interested in participating, add yourself to the issue and add to the PR #158.

(Admittedly, I have been occupied with millions of other things and it's a bit slow going but it's going to happen eventually and anyone who contributed to the package has a chance to be on the paper.)

PR #169, opened 2019-04-08
fixed xpm reader for py3

andrejberg:
Hi,
these are some minor changes I've done to let the XPM reader work with python3 for me.


orbeckst:
Many thanks @andrejberg for finding this bug. It _should_ have been already Py 3 but because we didn't have a test, it slipped through. I added some tests and made the reader read compressed files.

orbeckst:
Many thanks @andrejberg ! 👍

PR #174, opened 2019-04-26
don't call `logger` directly, fix #173 

### LINKED ISSUE #173: 'Logger' object is not callable
Now I'm trying to use it via jupyter-lab (Python 3.7.1), however, it does not work thus far.

```python3
import gromacs
itp = gromacs.fileformats.TOP("test.top")
```

A TypeError happens:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-eb4adf3f7a54> in <module>
----> 1 itp = gromacs.fileformats.TOP("test.top")

~/.local/pyenv/versions/anaconda3-2018.12/envs/md_analysis/lib/python3.7/site-packages/GromacsWrapper-0.7.0-py3.7.egg/gromacs/fileformats/top.py in __init__(self, fname)
     94 
     95         self.molecules = []
---> 96         self._parse(fname)
     97         self.molecules = tuple(self.molecules)
     98 

~/.local/pyenv/versions/anaconda3-2018.12/envs/md_analysis/lib/python3.7/site-packages/GromacsWrapper-0.7.0-py3.7.egg/gromacs/fileformats/top.py in _parse(self, fname)
    250                     assert particletype in ('A', 'S', 'V', 'D')
    251                     if particletype not in ('A',):
--> 252                         self.logger('warning: non-atom particletype: "{0:s}"'.format(line))
    253 
    254                     sig = float(fields[5+shift])

TypeError: 'Logger' object is not callable
```

Do I have to set specific log file name before beginning?

theavey:
This should fix #173 because `logger` was called directly instead of, for example, `logger.debug` or `logger.warning`.

This should possibly be tested for, and it might be good to see if the same issue occurs other places (a search for `logger(` might be sufficient).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@ richardjgowers, theavey, andrejberg, orbeckst
   (Issue #149)
 * fixed NDX writing (#152)
 * fixed Python 3 compatibility of XPM reader (PR #169)
+* fixed bug in calling logger in fileformats.top (issue #173)
 
 2018-08-08      0.7.0
 orbeckst, dldotson, kain88-de, ianmkenney
--- a/gromacs/fileformats/top.py
+++ b/gromacs/fileformats/top.py
@@ -234,7 +234,7 @@ class TOP(blocks.System):
 
                     '''
                     if len(fields) not in (6,7,8):
-                        self.logger('skipping atomtype line with neither 7 or 8 fields: \n {0:s}'.format(line))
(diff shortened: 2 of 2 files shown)


orbeckst:
@theavey thanks for the quick fix. I added an entry to CHANGES. Sorry for not getting to it sooner. Typically, just ping me with @orbeckst to get a response.

More tests would certainly be welcome. I just don't have very much time to do it. Contributions are _highly_ welcome!!

PR #184, opened 2020-07-28
func_name support for python 3.0 added in xvg.py

PolyachenkoYA:
gromacs/fileformats/xvg.py: func_name support for python 3.0 added


PolyachenkoYA:
done

PR #187, opened 2020-08-04
xvg.plot fix

PolyachenkoYA:
get_cmap takes not the list of colors in a map bit a name of a map which must be in the predefined set of names.
I think what you want here is register_cmap.

But ultimately,  me it works with register_cmap and doesn't work with get_cmap


whitead:
Thanks @PolyachenkoYA for your PR! Maybe instead of changing how this kwarg functions, you could add documentation to clarify and add a new one for your use case of specifying a custom cmap?

gromacs/fileformats/xvg.py
```diff
@@ -612,7 +613,7 @@ def plot(self, **kwargs):
 
         color = kwargs.pop('color', self.default_color_cycle)
         try:
-            cmap = matplotlib.cm.get_cmap(color)
+            cmap = matplotlib.cm.get_cmap(color) if(sys.version_info[0] < 3) else matplotlib.cm.register_cmap(color)
```
orbeckst, line 616: I think [cm.get_cmap()](https://matplotlib.org/3.3.1/api/cm_api.html#matplotlib.cm.get_cmap ) is just supposed to accept "viridis" or "hot" or any of the other internal color map names.

What is the difference between importing mpl under Python 2 vs Python 3? I don't understand why we need to look a the Python version to make a mpl issue work.

orbeckst (changes requested):
@PolyachenkoYA can you give an example that fails? I am trying to understand what the actual bug is.

PolyachenkoYA:
For me, `get_cmap` accepted "hot" and others for the py2 but not for the
py3. In the py3, `register_cmap` did that. I suppose they changed what
`get_cmap` does when switching from 2 to 3.

orbeckst:
PR #195 should solve this issue. I had a look and understood that all that needed changing was the exception handling that switches between cmaps and color cycles.

PR #205, opened 2021-07-22
extend testing

### LINKED ISSUE #202: test multiple GROMACS versions
In the switch from Travis CI to GitHub actions #200 we started with a minimal setup that does not have all the features as the previous tests. At the moment, only the latest version of GROMACS (2021, no mpi) is tested

Test against other available GROMACS versions in [bioconda/gromacs](https://anaconda.org/bioconda/gromacs/files). We don't have to test all of them always. Perhaps do a monthlt cron job with everything and test

- [ ] 2021 nompi
- [ ] 2021 mpi
- [ ] 2018
- [ ] 4.6.5

by default; on OSX 4.6.5 is not available.

## Linux
-  2021 mpi_openmpi
-  2020
-  2019
-  2018
-  4.6.5

2016 is not in the repository. 

## osx

-  2021 (nompi and mpi_openmpi)
-  2020
-  2019
-  2018

### LINKED ISSUE #203: test on MacOS
Since the switch of CI to GitHub actions #200 we are currently only testing on **Linux** runners. There are versions of GROMACS available for MacOS #202 so we should be running at least some tests (Python 2.7 and 3.8) on MacOS runners, too.

The GROMACS versions can be limited, e.g.
- [ ] Python 2.7, GROMACS 2021
- [ ] Python 3.8, GROMACS 2021

for a start.

orbeckst:
- fix #202 (test multiple GROMACS versions 4.6.5, 2018, 2020)
- fix #203 (test on macOS)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,16 +15,42 @@ concurrency:
   group: "${{ github.ref }}-${{ github.head_ref }}"
   cancel-in-progress: true
 
-# TODO: test different OS (but difficult with GROMACS versions)
-#       for right now: test on Linux only
 jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macOS-latest]
         python-version: [2.7, 3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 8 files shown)


orbeckst:
The test matrix is too big, need to only test all GROMACS versions with 2.7 and 3.8 (or 3.9) on Ubuntu, everything else can just be GMX 2021.

PR #270, opened 2023-11-11
fix #252: handle keyboard interrupts

### LINKED ISSUE #252: Handle KeyboardInterrupts when using `MDRunner.run()`
Sending a `Ctrl-C` interrupt during an mdrun kills the program but, as far as I can tell, leaves the simulation running with no easy way to kill it. I think that, ideally, interrupting the program should also kill background simulations.

jandom:
I have again paired with my coding robot and managed to hammer out some nice fix – plus a test!

### DIFF
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -5,7 +5,11 @@
 
 from __future__ import division, absolute_import, print_function
 
+import time
 import pytest
+import os
+import signal
+import threading
 
 from .datafiles import datafile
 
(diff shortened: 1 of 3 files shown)


tests/test_run.py
```diff
@@ -54,6 +58,49 @@ def test_MDRunner():
...
+    # Initialize MDrunner
+    mdrunner = gromacs.run.MDrunner()
+
+    # Function to send SIGINT after a delay
+    def send_interrupt():
+        time.sleep(1)  # Short delay before sending SIGINT
+        os.kill(os.getpid(), signal.SIGINT)
+
+    # Start a thread to send the SIGINT
+    interrupt_thread = threading.Thread(target=send_interrupt)
```
orbeckst, line 87: That is nifty/hacky. Did the robot suggest it?

orbeckst (changes requested):
Given that tests are happy, just add an entry to CHANGES.

You and your 🤖  are on a role. Maybe I have to update my priors regarding 🤖s.

jandom:
The robot did indeed suggest it but it took a lot of prompting :)

tests/test_run.py
jandom: The robot did indeed suggest it but it took a lot of prompting :)

jandom:
All changed now @orbeckst

orbeckst (approved):

PR #271, opened 2023-11-11
fix #263: No module named 'pkg_resources'

### LINKED ISSUE #263: Python 3.12: No module named 'pkg_resources'
```
  File "/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/gromacs/__init__.py", line 218, in <module>
    from . import config
  File "/opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/gromacs/config.py", line 266, in <module>
    from pkg_resources import resource_filename, resource_listdir
ModuleNotFoundError: No module named 'pkg_resources'
```

`pkg_resources` is a part of `setuptools`, and https://github.com/python/cpython/issues/95299 removes `setuptools` from Python environments. It needs to be manually added to the requirements.

jandom:
Just adding setuptools as a dep – will happily refactor in another PR to use importlib?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
         "Topic :: Scientific/Engineering :: Bio-Informatics",
         "Topic :: Scientific/Engineering :: Chemistry",
         "Topic :: Software Development :: Libraries :: Python Modules",
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ setup(
         "matplotlib",
         "numkit",  # numerical helpers
         "numpy>=1.0",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -66,6 +66,7 @@
         "matplotlib",
         "numkit",  # numerical helpers
         "numpy>=1.0",
+        "setuptools",  # contains 'pkg_resources' for versioneer
```
njzjz, line 69: Only require setuptools for Python>=3.12:
```suggestion
        'setuptools; python_version >= "3.12"',  # contains 'pkg_resources' for versioneer
```

orbeckst: Thanks @njzjz , I added you to AUTHORS.

orbeckst (approved):
Thanks for the quick fix & thank you for @njzjz for the input! (I added @njzjz to AUTHORS/CHANGES).

I also added CI for py 3.12 so that we can now officially claim to support py 3.12.

PR #272, opened 2023-11-11
ci: try a simple ci.yaml (experiment)

jandom:
When setting up to work on this repo, I found the setup very challenging. There is no setup script, so I worked based on what the ci.yaml does – and that contained a number of curveballs. Luckily things can be simplified and everything still works!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,8 +81,6 @@ jobs:
         create-args: >-
            python=${{ matrix.python-version }}.*=*_cpython
            gromacs==${{ matrix.gromacs-version }}
-    
-    - run: micromamba install pocl
 
     - name: Python version information ${{ matrix.python-version }}
       run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,6 +81,8 @@ jobs:
         create-args: >-
            python=${{ matrix.python-version }}.*=*_cpython
            gromacs==${{ matrix.gromacs-version }}
+    
+    - run: micromamba install pocl
 
     - name: Python version information ${{ matrix.python-version }}
       run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,7 +70,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: micromamba environment and package installation (Python ${{ matrix.python-version }})
+    - name: micromamba package and testing environment installation (Python ${{ matrix.python-version }}, GROMACS ${{ matrix.gromacs-version }})
       uses: mamba-org/setup-micromamba@v1
       with:
         environment-file: ci/conda-envs/test_env.yaml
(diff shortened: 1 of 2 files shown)


jandom:
@orbeckst this one puts a smile on my face and makes things **a bit** simpler

.github/workflows/ci.yaml
```diff
@@ -94,17 +95,6 @@ jobs:
          micromamba list
          cat /proc/cpuinfo || (/usr/sbin/system_profiler SPHardwareDataType; /usr/sbin/sysctl -a | grep machdep.cpu)
 
-    - name: Install pytest and plugins
-      run: |
-         micromamba install pytest pytest-pep8 pytest-cov codecov
-
-    - name: Install GROMACS (${{ matrix.gromacs-version }})
```
orbeckst, line 101: update `name: micromamba environment and package installation (Python ${{ matrix.python-version }})` to also include GROMACS version, along the lines of 

```yaml
name: micromamba package and testing environment installation (Python ${{ matrix.python-version }}, GROMACS ${{ matrix.gromacs-version }})

File: .github/workflows/ci.yaml [hunk 2]
```diff
@@ -94,17 +95,6 @@ jobs:
...
-      # UGLY HACK/FIX (probably to issues with bioconda packages)
-      # - For 3.9, micromamba insists on using pypy 3.9 (and --py-pin does not help).
-      # We can't freeze because then it's not possible to install older GROMACS versions from bioconda.
```
orbeckst, line 104: I really hope that this problem doesn't come back in any form because hacking around this dependency hell cost me a lot of sweat and gray hairs.

orbeckst: Add a comment above somewhere near where the env.yml file is defined that it is important to install gromacs together with everything else to ensure correct dependency resolution.

orbeckst (changes requested):
To summarize that I get it: The new idea is to install the whole testing environment in one go (with the appropriate Python and GROMACS versions and pytest) and in this way 
1. have a single env.yml file that one can also use conveniently for local testing
2. avoid the hacky install of GROMACS that forces the installation of the required python because of the weird/broken dependency resolution around PyPy python 3.9

This looks ok to me. Generally, I like splitting up tasks and debug and monitor separately. But I can get behind 1 and 2 above.

I have some small comments about naming.

The big thing is: Please go through the actions and double check that the correct versions of GROMACS and Python are actually installed. You need to look at the final `mamba list` to see what's _actually_ in the build environment. Then confirm that this is all correct for the desired entries in our build matrix.

.github/workflows/ci.yaml
jandom: good catch, thanks – updated

.github/workflows/ci.yaml
jandom: agreed – i've added some grep checks to confirm that what's been requested has been installed

orbeckst (approved):

orbeckst:
thanks!

jandom:
Thank you for the feedback 🎉

PR #273, opened 2023-11-13
refactor: use importlib instead of pkg_resources

jandom:
Removing the setuptools dep, use importlib instead

### DIFF
--- a/gromacs/config.py
+++ b/gromacs/config.py
@@ -232,7 +232,7 @@ import subprocess
 
 from configparser import ConfigParser
 
-from pathlib import Path
+from importlib import resources
 
 from . import utilities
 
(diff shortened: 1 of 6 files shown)


jandom:
Not using importlib at all – and i'm not sure this will work outside of CI/CD?

gromacs/config.py
```diff
@@ -309,23 +308,15 @@
...
-    )
-
-
-def resource_basename(resource):
-    """Last component of a resource (which always uses '/' as sep)."""
-    if resource.endswith("/"):
-        resource = resource[:-1]
-    parts = resource.split("/")
-    return parts[-1]
+    resource_path = Path(__package__) / dirname
```
orbeckst, line 314: This does not necessarily work with installed packages, I think, or at least if these are zipped or eggs or whatever.

https://importlib-resources.readthedocs.io/en/latest/using.html#file-system-or-zip-file and in particular the `atexit` solution in https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-filename

Probably needs something like
```python
import importlib.resources as importlib_resources
from contextlib import ExitStack
import atexit

file_manager = ExitStack()
atexit.register(file_manager.close)
resource_path = importlib_resources.files(gromacs.config) / dirname
path = file_manager.enter_context(importlib_resources.as_file(resource_path))
```

orbeckst (changes requested):
Probably needs a more sophisticated version, see comment.

jandom:
This just looks hard and I'm not enjoying it too much – let's work on sth else, and return to this one with a fresh mind later

orbeckst:
That’s fair ;-)

PR #275, opened 2024-05-16
fix TestAttributeDict

njzjz:
`setup` -> `setup_method`


njzjz:
Ok, it works. Two failed tests are unrelated.

orbeckst (changes requested):
Thanks for the fix. I didn't know (or had forgotten) that I could have just changed the name of the method. 

I have independently fixed it in PR #278 and use a proper fixture. 

I'll close this PR but many thanks for putting in the effort!!!! 🙏

PR #284, opened 2024-06-16
use importlib.resources instead of pkg_resources

### LINKED ISSUE #282: replace pkg_resources everywhere
We use `pkg_resources` to pull templates out of the installed package but [pkg_resources is deprecated as an API](https://setuptools.pypa.io/en/latest/pkg_resources.html) and should be replaced with code based on [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources).

These are the two occurences that are left:
* gromacs/config.py:from pkg_resources import resource_filename, resource_listdir
* tests/datafiles.py:from pkg_resources import resource_filename

orbeckst:
fix #282

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,7 +3,13 @@
 ==============================
 
 2024-??-??      0.9.1
+orbeckst
 
+* templates in `config.templates` are now stored as
+  `importlib.resources.abc.Traversable` (PosixPath-like) objects instead of
+  strings of paths to files in the file system because we switched from
+  pkg_resources to importlib.resources for managing access to included files
+  (#282)
 * fixed reporting of failures of GromacsCommand and DeprecationWarnings for use
   of \s in regular expression strings (#285)
 
--- a/gromacs/config.py
+++ b/gromacs/config.py
@@ -232,8 +232,7 @@ import subprocess
 import sys
 
 from configparser import ConfigParser
-
-from pkg_resources import resource_filename, resource_listdir
+from importlib import resources
 
 from . import utilities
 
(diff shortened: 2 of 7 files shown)


orbeckst:
3.9 - 3.11 runners fail with errors such as
```
==================================== ERRORS ====================================
______________ ERROR collecting tests/fileformats/test_convert.py ______________
tests/fileformats/test_convert.py:12: in <module>
    from gromacs.fileformats import convert
gromacs/__init__.py:216: in <module>
    from . import config
gromacs/config.py:328: in <module>
    templates = _generate_template_dict("templates")
gromacs/config.py:324: in _generate_template_dict
    for entry in resources.files(__name__).joinpath(dirname).iterdir()
../../../micromamba/envs/gromacs_wrapper/lib/python3.11/importlib/resources/_common.py:22: in files
    return from_package(get_package(package))
../../../micromamba/envs/gromacs_wrapper/lib/python3.11/importlib/resources/_common.py:55: in get_package
    raise TypeError(f'{package!r} is not a package')
E   TypeError: 'gromacs.config' is not a package
```

Python 3.12 works.

orbeckst:
In 3.12, we can use a module instead of a package as an *anchor* for [importlib.resources.files](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files)

> Changed in version 3.12: package parameter was renamed to anchor. anchor can now be a non-package module and if omitted will default to the caller’s module. package is still accepted for compatibility but will raise a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). Consider passing the anchor positionally or using importlib_resources >= 5.10 for a compatible interface on older Pythons.

PR #286, opened 2024-06-16
fix "DeprecationWarning: invalid escape sequence '\s'"

### LINKED ISSUE #285: DeprecationWarning: invalid escape sequence '\s'
fix up all regular expressions where we use `\s`.

orbeckst:
fix #285

### DIFF
--- a/gromacs/core.py
+++ b/gromacs/core.py
@@ -434,8 +434,8 @@ class GromacsCommand(Command):
 
     command_name = None
     driver = None
-    doc_pattern = """.*?(?P<DOCS>DESCRIPTION.*)"""
-    gmxfatal_pattern = """----+\n                   # ---- decorator line
+    doc_pattern = r""".*?(?P<DOCS>DESCRIPTION.*)"""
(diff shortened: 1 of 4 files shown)


gromacs/core.py
```diff
@@ -434,8 +434,8 @@ class GromacsCommand(Command):
 
     command_name = None
     driver = None
-    doc_pattern = """.*?(?P<DOCS>DESCRIPTION.*)"""
-    gmxfatal_pattern = """----+\n                   # ---- decorator line
+    doc_pattern = r""".*?(?P<DOCS>DESCRIPTION.*)"""
+    gmxfatal_pattern = r"""----+\n                  # ---- decorator line
```
orbeckst, line 438: I am not 100% sure that the newline is actually captured by the raw-string expression.

Ought to check if there's _actually_ a test for that.

orbeckst: Now there is.

(Of course it hadn't worked ... untested code is broken code!)

PR #293, opened 2024-09-13
configure codecov

### LINKED ISSUE #287: codecov coverage dropped to 0%
The codecov coverage for PRs shows 0% and the total coverage for main is also 0% even though it should be around 69%.

I recently updated [.codecov.yml](https://github.com/Becksteinlab/GromacsWrapper/blob/main/.codecov.yml) to point to *main* so maybe that threw something off.

Or maybe it will all go away by itself... 🌈 🦄

orbeckst:
- fix #287 
- split project status into package and tests
- split patch status, too

Note: check codecov.yaml with
```
curl --data-binary @codecov.yml https://codecov.io/validate
```


orbeckst (commented):
Not sure if this really fixes #287 , at least not in the PR. I will merge and re-open the issue if this does not do anything.

PR #294, opened 2024-09-14
testing codecov

orbeckst:
Test different versions of `codecov.yml` to see if anything solves #287 

Validate with
```
curl --data-binary @codecov.yml https://codecov.io/validate
```


orbeckst:
I was able to get normal coverage when running locally
```
pytest -v --disable-pytest-warnings --durations=20 --low-performance  --cov-report=xml  --cov-report=term-missing --color=yes ./tests
coverage report
```

I can locally reproduce the *No data to report.* --- I think after adding the `pytest-pep8` plugin to pytest-cov (and pytest-xdist) (python 3.13).

When removing the plugin (`mamba uninstall pytest-pep8`) the lack of coverage remains, so maybe something else changed.

Issue #298, opened 2024-09-30
Cron CI failure: timeout in micromamba list


orbeckst:
The latest CI cron failed https://github.com/Becksteinlab/GromacsWrapper/actions/runs/11081172277

> The job running on runner GitHub Actions xxx has exceeded the maximum execution time of 360 minutes.

It seems to be the 
```
micromamba list
```
step.

Need to investigate.

orbeckst:
Possibly https://github.com/mamba-org/setup-micromamba/issues/225

orbeckst:
Note that our CI actually wants to install setup-micromamba@v1 whereas the linked issue talks about v2.0.0 as far as I can tell. There's some chatter that GH actions uses its own micromamba and ignores the installed one but I don't see this in our logs, eg https://github.com/Becksteinlab/GromacsWrapper/actions/runs/11081172277/job/30886588889 : 

1. Install to /home/runner/micromamba-bin/micromamba
```
Run mamba-org/setup-micromamba@v1
  with:
    environment-file: ci/conda-envs/test_env.yaml
    condarc: channels:
     - conda-forge
     - bioconda
  channel_priority: flexible         
  
    cache-downloads: true
    cache-environment: true
    create-args: python=3.11.*=*_cpython gromacs==2018.6
  env:
    MPLBACKEND: agg
Will download micromamba to /home/runner/micromamba-bin/micromamba
Install micromamba
```
Curiously, we **do get 2.0.1**
```
 micromamba version : 2.0.1
```
2. run list
```
/home/runner/micromamba-bin/micromamba list -r /home/runner/micromamba -n gromacs_wrapper
```

orbeckst:
I just deleted **all caches** for GH actions and now trying to rerun.

orbeckst:
Deleting caches did not help.

PR #300, opened 2024-10-01
comment out micromamba list

orbeckst:
temp fix for #298


orbeckst:
This does not help. The problem is inside the micromamba-setup action.

